PR #2, opened 2022-09-13
Stabilize the testcases

mprahl:
See each commit for the description of each change.


JustinKuli (approved):
Why is patch better than update? I don't have a problem with it, and I do like it in a separate function. I'm just curious.

JustinKuli:
I bet this is a flakey test: https://github.com/open-cluster-management-io/governance-policy-framework-addon/actions/runs/3053002798/jobs/4923066818#step:10:6892.

I wonder if we should change the fail-fast strategy in the matrix, like in https://github.com/stolostron/governance-policy-framework/blob/main/.github/workflows/kind.yml#L22-L25

mprahl:
> Why is patch better than update? I don't have a problem with it, and I do like it in a separate function. I'm just curious.

The issue with update is you need to get the object from the API, make changes, and resubmit it. If there was a change made to that object since you got the object from the API, the update fails. A patch doesn't have this same limitation and it's cleaner than retries.

mprahl:
> I bet this is a flakey test: https://github.com/open-cluster-management-io/governance-policy-framework-addon/actions/runs/3053002798/jobs/4923066818#step:10:6892.

I'll try to address this in the PR. I'm having trouble reproducing this locally.

> I wonder if we should change the fail-fast strategy in the matrix, like in https://github.com/stolostron/governance-policy-framework/blob/main/.github/workflows/kind.yml#L22-L25

I think there's value to that, but I don't have a strong preference.

dhaiducek:
Neat. Well, this time it passed... 🤷

gparvin (approved):

PR #8, opened 2022-10-20
Add dynamic watches for dependencies

JustinKuli:
Template-sync's Reconcile function will now be called for a policy whenever that policy is changed, or when any of its dependencies change.

Refs:
 - https://github.com/stolostron/backlog/issues/26186

Signed-off-by: Justin Kulikauskas <jkulikau@redhat.com>


controllers/templatesync/template_sync.go
```diff
@@ -121,6 +126,19 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 		return reconcile.Result{}, nil
 	}
 
+	allDeps := make(map[depclient.ObjectIdentifier]bool)
```
willkutler, line 129: Optional - it would be a bit cleaner to use something like https://github.com/deckarep/golang-set since we don't use the bool value, but up to you if you don't want to import another library

willkutler:
@JustinKuli this looks good to me! Added an optional comment and I think the imports need some formatting but otherwise it looks good to go

controllers/templatesync/template_sync.go
JustinKuli: I don't know if there's enough advantage to using another library for this, I think it would only clean up the conversion to a slice? If we had more set-specific operations like intersections or joins, then I'd be happy to use the library.

main.go
```diff
@@ -352,12 +353,33 @@ func getManager(
...
+
+	go func() {
+		err := watcher.Start(mgrCtx)
+		if err != nil {
+			panic(err)
+		}
+	}()
+
+	// Wait until the dynamic watcher has started.
+	<-watcher.Started()
```
mprahl, line 380: You could move this to right before `return mgr` so that the rest of the setup can continue while dynamic watcher is starting.

mprahl (approved):

PR #11, opened 2022-10-31
Set pending state on policies if dependencies are not met + remove template

willkutler:
Adds logic to determine whether a policy template should be pending based on whether its dependencies are satisfied. If a template should be pending, these changes will remove it from the managed cluster and set the status of the parent policy to Pending.

https://github.com/stolostron/backlog/issues/26188
https://github.com/stolostron/backlog/issues/26189

### DIFF
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/go-logr/logr"
 	depclient "github.com/stolostron/kubernetes-dependency-watches/client"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -61,10 +62,11 @@ type PolicyReconciler struct {
 	// This client, initialized using mgr.Client() above, is a split client
 	// that reads objects from the cache and writes to the apiserver
 	client.Client
-	DynamicWatcher depclient.DynamicWatcher
-	Scheme         *runtime.Scheme
-	Config         *rest.Config
-	Recorder       record.EventRecorder
+	DynamicWatcher   depclient.DynamicWatcher
+	Scheme           *runtime.Scheme
+	Config           *rest.Config
+	Recorder         record.EventRecorder
+	ClusterNamespace string
 }
 
 // Reconcile reads that state of the cluster for a Policy object and makes changes based on the state read
@@ -126,7 +128,8 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 		return reconcile.Result{}, nil
 	}
 
-	allDeps := make(map[depclient.ObjectIdentifier]bool)
+	allDeps := make(map[depclient.ObjectIdentifier]string)
+	topLevelDeps := make(map[depclient.ObjectIdentifier]string)
 
 	for _, dep := range instance.Spec.Dependencies {
 		depID := depclient.ObjectIdentifier{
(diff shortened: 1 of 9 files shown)


test/e2e/case12_ordering_test.go
```diff
@@ -0,0 +1,385 @@
...
+			"policy: managed/namespace-foo-setup-configpolicy",
+			"NonCompliant; there is violation")
+
+		By("Checking if dependency status is noncompliant")
+		Eventually(checkCompliance(case12DepName), defaultTimeoutSeconds, 1).Should(Equal("NonCompliant"))
+
+		By("Checking if policy status is pending")
+		Eventually(checkCompliance(case12PolicyName), defaultTimeoutSeconds, 1).Should(Equal("Pending"))
+	})
+	It("Should set to Compliant when dep status is resolved", func() {
```
JustinKuli, line 148: Unnecessary suggestion: This test begins by repeating what the previous test did. If the `Ordered` keyword was added to this parent `Describe`, and the `After/BeforeEach` were converted to `After/BeforeAll`, then the repetition would not be needed.

But that might require other test re-organization, so it's not necessary.

test/e2e/case12_ordering_test.go
```diff
@@ -0,0 +1,385 @@
...
+				case12PolicyName,
+				clusterNamespace,
+				true,
+				defaultTimeoutSeconds)
+
+			return getCompliant(managedPlc)
+		}, defaultTimeoutSeconds, 1).Should(Equal("Compliant"))
+	})
+	It("Should remove template if dependency changes", func() {
+		By("Generating an compliant event on the dependency")
```
JustinKuli, line 224: Nit
```suggestion
		By("Generating a compliant event on the dependency")
```

there are a few spots like this

test/e2e/case12_ordering_test.go
```diff
@@ -0,0 +1,385 @@
...
+			managedPlc = utils.GetWithTimeout(
+				clientManagedDynamic,
+				gvrPolicy,
+				case12DepName,
+				clusterNamespace,
+				true,
+				defaultTimeoutSeconds)
+
+			return getCompliant(managedPlc)
+		}, defaultTimeoutSeconds, 1).Should(Equal("Compliant"))
```
JustinKuli, line 354: Does the "short version" you use in other places work here?
```
		By("Checking if dependency status is compliant")
		Eventually(checkCompliance(case12DepBName), defaultTimeoutSeconds, 1).Should(Equal("Compliant"))
```

Or, if there's a difference I'm missing, can that be put into another new function? I just see a lot of repetition of multiple lines, and it makes it harder for me to check what the test is really doing.

test/e2e/case12_ordering_test.go
```diff
@@ -0,0 +1,385 @@
...
+			case12PolicyName,
+			clusterNamespace,
+			true,
+			defaultTimeoutSeconds)
+		Expect(managedPlc).NotTo(BeNil())
+		managedRecorder.Event(
+			managedPlc,
+			"Normal",
+			"policy: managed/case12-config-policy",
+			"Compliant; No violation detected")
```
JustinKuli, line 368: Nit: this would be another candidate (in my mind) for a helper function to reduce repetition.

JustinKuli (changes requested):
Looking good overall! I have a lot of somewhat optional comments, and only one that I think actually would require a change: whether we can use `allDeps`, or whether we need a `tmplDeps` variable.

controllers/templatesync/template_sync.go
```diff
@@ -136,7 +137,17 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
-		allDeps[depID] = true
+
+		existingDep, ok := allDeps[depID]
+		if ok && existingDep != dep.Compliance {
+			err := fmt.Errorf("dependency on %s has conflicting compliance states", dep.Name)
+
+			reqLogger.Error(err, "Failed to decode the policy dependencies", "policy", instance.GetName())
+
+			continue
+		}
```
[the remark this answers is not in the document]
willkutler: yeah, that's right. It should also catch conflicts between deps and extra deps

test/e2e/case12_ordering_test.go
willkutler: I removed the before/afterAll because other tests don't use the same policies, so I think this isn't needed anymore

JustinKuli (approved):
/hold

Looks good to me! Just holding to squash the commits (which helped in review, so thanks for keeping them like that), and I think it will be good to go.

JustinKuli (approved):

willkutler:
/unhold

commits have been squashed and the message has been updated, this PR should be good to go now

PR #13, opened 2022-11-08
Fix a bug where the /healthz endpoint pointed to underlying /readyz

mprahl:
This bug caused the /healthz endpoint on the proxy to combine the controller
manager's /readyz endpoints. This would lead to the controller not
restarting when the Hub kubeconfig changed.

Relates:
https://issues.redhat.com/browse/ACM-1932

### DIFF
--- a/main.go
+++ b/main.go
@@ -505,6 +505,8 @@ func startHealthProxy(ctx context.Context, wg *sync.WaitGroup, addresses ...stri
 	log := ctrl.Log.WithName("healthproxy")
 
 	for _, endpoint := range []string{"/healthz", "/readyz"} {
+		endpoint := endpoint
+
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -526,9 +529,7 @@ func startHealthProxy(ctx context.Context, wg *sync.WaitGroup, addresses ...stri
 				defer resp.Body.Close()
 
 				if resp.StatusCode != http.StatusOK {
-					body := []byte{}
-
-					_, err = resp.Body.Read(body)
+					body, err := ioutil.ReadAll(resp.Body)
```
JustinKuli, line 532: Can this use `io.ReadAll` instead? https://pkg.go.dev/io/ioutil#ReadAll is deprecated

JustinKuli (changes requested):

PR #15, opened 2022-11-15
Use cluster namespace for known policy types

JustinKuli:
Policies and ConfigurationPolicies should all exist in the cluster namespace, so setting a dependency with one of those types that uses another namespace is generally incorrect. The main use of the namespace field is to identify root policies in other namespaces, but that is handled in the propagator when the dependencies are canonicalized.

This keeps open the possibility of defining dependency on user-created policy types which won't necessarily follow this rule, and could have a meaningful namespace.

Signed-off-by: Justin Kulikauskas <jkulikau@redhat.com>

### DIFF
--- a/.github/workflows/kind.yml
+++ b/.github/workflows/kind.yml
@@ -21,6 +21,7 @@ jobs:
     env:
       REGISTRY: localhost:5000
     strategy:
+      fail-fast: false
       matrix:
         # Run tests on oldest and newest supported OCP Kubernetes
         # - OCP 4.5 runs Kubernetes v1.18
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -518,9 +518,18 @@ func (r *PolicyReconciler) processDependencies(ctx context.Context, dClient dyna
 			return nil, err
 		}
 
-		// set up namespace for replicated policy dependencies
 		ns := dep.Namespace
+
+		// infer namespace if not provided
 		if ns == "" {
+			ns = r.ClusterNamespace
+		}
+
+		// override given namespace if one of our policy types
+		// (these will always be in the cluster namespace)
+		if dep.Group == policiesv1.GroupVersion.Group &&
+			dep.Version == policiesv1.GroupVersion.Version &&
+			strings.HasSuffix(dep.Kind, "Policy") {
 			ns = r.ClusterNamespace
 		}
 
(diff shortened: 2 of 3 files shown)


controllers/templatesync/template_sync.go
```diff
@@ -518,12 +518,21 @@ func (r *PolicyReconciler) processDependencies(ctx context.Context, dClient dyna
...
 		if ns == "" {
 			ns = r.ClusterNamespace
 		}
 
+		// override given namespace if one of our policy types
+		// (these will always be in the cluster namespace)
+		if dep.Group == policiesv1.GroupVersion.Group &&
+			dep.Version == policiesv1.GroupVersion.Version &&
+			strings.HasSuffix(dep.Kind, "Policy") {
+			ns = r.ClusterNamespace
```
dhaiducek, line 533: I'm trying to decide whether this should be an error case? It seems like we don't necessarily want to assume that the user should be overridden, but maybe I'm missing some implementation details here.

JustinKuli: It could be an error. I was trying to go for ease-of-use here.

Example: I want my policy to be active when a separate ConfigurationPolicy is Compliant. What namespace should I use for the dependency? I might think that I should use the namespace of the root Policy containing the ConfigurationPolicy, but actually that namespace only exists on the hub. On the managed cluster, the ConfigurationPolicy exists in the managed cluster namespace, but I can't put that in the dependency definition because each cluster is different.

So this makes the framework always use the cluster namespace, instead of requiring the user to leave the namespace blank.

dhaiducek: It makes sense for ease of use, but I suppose I don't like that it could propagate misconceptions if the user puts in a namespace they think is correct and we override it without letting them know it's actually wrong. I think I'd prefer an informative message that they need to remove or update the namespace.

.github/workflows/kind.yml
```diff
@@ -21,6 +21,7 @@ jobs:
     env:
       REGISTRY: localhost:5000
     strategy:
+      fail-fast: false
```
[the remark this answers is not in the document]
dhaiducek: Makes sense! My concern there is speed, but it's easy enough to modify the Makefile to run a subset of tests if something keeps failing consistently because of the code in the PR.

controllers/templatesync/template_sync.go
JustinKuli: Yeah, "propagating misconceptions" definitely sounds like a bad thing. You've convinced me! I'll look at putting a warning/error in the propagator logic for this situation instead of overriding it here.

PR #21, opened 2022-11-30
Handle ignorePending differently

JustinKuli:
Previously, individual templates were always marked as "Pending" if they
had an unsatisfied dependency. The `ignorePending` field was only
considered when calculating the overall compliance of a Policy. But that
meant that it was not possible to correctly calculate the overall state
with just the status object, which is confusing.

Now, the `ignorePending` field is considered when marking the compliance
of each template. This makes how the overall compliance is calculated
more clear, and the new messages may be more helpful for users.

Refs:
- https://issues.redhat.com/browse/ACM-2101


willkutler (approved):

willkutler:
Looks good! Thanks for the changes @JustinKuli

PR #23, opened 2022-12-09
Add template-sync metrics

dhaiducek:
Adds metrics to track user vs system errors.

ref: https://issues.redhat.com/browse/ACM-2217


test/utils/common.go
```diff
@@ -23,3 +26,32 @@ func CreateRecorder(kubeClient kubernetes.Interface, componentName string) (reco
 
 	return eventBroadcaster.NewRecorder(eventsScheme, v1.EventSource{Component: componentName}), nil
 }
+
+// getMetrics curls the metrics endpoint, filters the response with the given patterns,
+// and returns the value(s) for the matching metric(s).
+func GetMetrics(metricPatterns ...string) []string {
```
JustinKuli, line 32: Looks familiar 😂

controllers/templatesync/template_metric.go
```diff
@@ -0,0 +1,49 @@
...
+var (
+	policyUserErrorsCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "policy_user_errors_total",
+			Help: "The number of user errors encountered while processing policies",
+		},
+		[]string{
+			"policy",
+			"template",
+			"type",
```
JustinKuli, line 19: I see what you were going for with the `type` label, but I've started getting more paranoid about the number of time-series we're exporting. Do you think it's required, or useful enough to be worth the possible performance concerns? (Performance concerns on the prometheus instance, not here in our controller.)

I feel like we could skip that label, and grep logs for counts of specific kinds of errors when we need to look closer.

JustinKuli: The Openshift Logging docs didn't help me much, but this blog has a nice view on how it could be used (jump to "The Openshift logging operator"): https://myopenshiftblog.com/openshift-observability/

dhaiducek: Good point. I was following patterns in the ConfigPolicy PR: https://github.com/open-cluster-management-io/config-policy-controller/pull/84

Though `type` is also a constrained label since we'd have a finite number of types. The `Policy` and `Template` labels are arguably more problematic.

I suppose the question here is whether SD is going to use Openshift Logging (which they should) or whether we should do this for now. I'm not actually sure where we came up with these labels or if the first implementation used them and we've been using them everywhere else ever since...

JustinKuli: Matching the previous work seems fine. Just something to watch out for.

JustinKuli (approved):

PR #28, opened 2023-02-10
Clean up policy templates that have been removed from the parent policy

willkutler:
Adds a label to policy templates that are created as a child of a policy, then compares the policy templates actually contained in that policy to all templates with the parent policy label to determine if templates have been renamed or removed from the parent. The template sync will then clean up any policy templates that are not being managed by the parent policy that created them.

refs:
- https://issues.redhat.com/browse/ACM-3049?filter=-1

Signed-off-by: Will Kutler <wkutler@redhat.com>

### DIFF
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -12,11 +12,14 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/prometheus/client_golang/prometheus"
 	depclient "github.com/stolostron/kubernetes-dependency-watches/client"
+	extensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	extensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
@@ -36,14 +39,17 @@ import (
 )
 
 const (
-	ControllerName string = "policy-template-sync"
-	policyFmtStr   string = "policy: %s/%s"
+	ControllerName    string = "policy-template-sync"
+	policyFmtStr      string = "policy: %s/%s"
+	PolicyTypeLabel   string = "policy.open-cluster-management.io/policy-type"
+	parentPolicyLabel string = "policy.open-cluster-management.io/policy"
 )
 
 var log = ctrl.Log.WithName(ControllerName)
 
 //+kubebuilder:rbac:groups=policy.open-cluster-management.io,resources=*,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=core,resources=events,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list
 
 // Setup sets up the controller
 func (r *PolicyReconciler) Setup(mgr ctrl.Manager, depEvents *source.Channel) error {
@@ -177,6 +183,8 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 	// As a quirk of the error handling, only the last occurring error is "returned" by Reconcile.
 	var resultError error
 
+	var templateNames []string
+
 	// PolicyTemplates is not empty
 	// loop through policy templates
 	for tIndex, policyT := range instance.Spec.PolicyTemplates {
@@ -258,6 +266,8 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 
 			continue
 		}
+
+		templateNames = append(templateNames, tName)
 
 		tLogger := reqLogger.WithValues("template", tName)
 
@@ -354,6 +364,9 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 					labels["cluster-namespace"] = instance.GetLabels()[common.ClusterNamespaceLabel]
 					labels[common.ClusterNamespaceLabel] = instance.GetLabels()[common.ClusterNamespaceLabel]
 				}
+
+				// set label to identify parent policy for this template
+				labels[parentPolicyLabel] = instance.GetName()
 
 				tObjectUnstructured.SetLabels(labels)
 				tObjectUnstructured.SetOwnerReferences([]metav1.OwnerReference{plcOwnerReferences})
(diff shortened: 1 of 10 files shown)


willkutler:
ready for review!

controllers/templatesync/template_sync.go
```diff
@@ -523,11 +531,89 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
+					found = true
+				}
+			}
+
+			if !found {
+				err := dClient.Resource(schema.GroupVersionResource{
+					Group:    "policy.open-cluster-management.io",
+					Version:  "v1",
+					Resource: "configurationpolicies",
+				}).Namespace(r.ClusterNamespace).Delete(ctx, tmpl.GetName(), metav1.DeleteOptions{})
```
JustinKuli, line 606: This should be from `gvr`, not always `configurationpolicy`.

JustinKuli (changes requested):
I think this could have a problem if I have two templates of different `kinds`, but the same name? So instead of just names of the templates, maybe it should store kind+name.

controllers/templatesync/template_sync.go
```diff
@@ -44,6 +44,7 @@ var log = ctrl.Log.WithName(ControllerName)
 
 //+kubebuilder:rbac:groups=policy.open-cluster-management.io,resources=*,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=core,resources=events,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list
```
mprahl, line 47: Thanks for remembering this!

controllers/templatesync/template_sync.go
```diff
@@ -523,11 +531,89 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
+// cleanUpExcessTemplates compares existing policy templates on the cluster to those contained in the policy,
+// and deletes those that have been renamed or removed from the parent policy
+func (r *PolicyReconciler) cleanUpExcessTemplates(
+	ctx context.Context,
+	dClient dynamic.Interface,
+	instance policiesv1.Policy,
+	parentTemplates []string,
+) error {
+	// build list of policy template resources to check
+	tmplCRDs, err := dClient.Resource(schema.GroupVersionResource{
```
mprahl, line 554: Could you please take advantage of caching by using the client associated with the `PolicyReconciler`? Additionally, could you please configure the cache with the label selector similar to what the config-policy-controller does?

https://github.com/open-cluster-management-io/config-policy-controller/blob/8203f5fb9e5fbd72b85852d54f7bd91331c298ce/main.go#L180-L187

controllers/templatesync/template_sync.go
```diff
@@ -523,11 +531,89 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
+	tmplGVRs := []schema.GroupVersionResource{}
+
+	for _, crd := range tmplCRDs.Items {
+		versions, ok := crd.Object["spec"].(map[string]interface{})["versions"].([]interface{})
+		if !ok {
+			return fmt.Errorf("template CRD is formatted incorrectly: no versions found")
+		}
+
+		for _, version := range versions {
+			tmplGVRs = append(tmplGVRs, schema.GroupVersionResource{
```
mprahl, line 574: You don't need to select every version. You just need to pick one that is supported by the CRD (i.e. the latest one).

controllers/templatesync/template_sync.go
```diff
@@ -523,11 +531,89 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
+			return fmt.Errorf("error listing %s objects: %w", gvr.String(), err)
+		}
+
+		for _, tmpl := range children.Items {
+			// delete all templates with label that aren't still in the policy
+			found := false
+
+			for _, parentTmplName := range parentTemplates {
+				if parentTmplName == tmpl.GetName() {
+					found = true
```
mprahl, line 597: Can you break after this statement to exit the for loop?

test/e2e/case15_template_cleanup_test.go
```diff
@@ -0,0 +1,111 @@
...
+		Expect(err).Should(BeNil())
+		plc := utils.GetWithTimeout(clientManagedDynamic, gvrPolicy, case15PolicyName, clusterNamespace, true,
+			defaultTimeoutSeconds)
+		Expect(plc).NotTo(BeNil())
+	})
+	AfterEach(func() {
+		By("Deleting a policy on the hub in ns:" + clusterNamespaceOnHub)
+		_, err := kubectlHub("delete", "-f", case15PolicyYaml, "-n", clusterNamespaceOnHub)
+		var e *exec.ExitError
+		if !errors.As(err, &e) {
```
mprahl, line 39: Could you use `--ignore-not-found=true` instead of this guard?

controllers/templatesync/template_sync.go
willkutler: @mprahl I'm working on this but am a bit unfamiliar with using the cache. I'm confused on choosing between v1 and v1beta1 - it seems like the cache selectors function you linked automatically chooses the right one, but when I use the reconciler client to list the CRDs, the controller needs to pick between v1 and v1beta1 for the object list it is populating with the CRDs, right? Like this:

```
	crds := extensionsv1.CustomResourceDefinitionList{}
	label, err := labels.NewRequirement(
		"policy.open-cluster-management.io/policy-type",
		selection.Equals,
		[]string{"template"},
	)
	err = r.List(ctx, &crds, &client.ListOptions{
		LabelSelector: labels.NewSelector().Add(*label),
	})
```

Unless there is a data type I'm not aware of that accepts a list of either v1 or v1beta1 CRDs, I'm not sure how to use the reconciler client here without a duplicate block that lists v1beta1 CRDs if the above block fails.

mprahl: Here is roughly what I had in mind:
```diff
diff --git a/controllers/templatesync/template_sync.go b/controllers/templatesync/template_sync.go
index 8036214..85d822e 100644
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -12,11 +12,14 @@ import (
        "github.com/go-logr/logr"
        "github.com/prometheus/client_golang/prometheus"
        depclient "github.com/stolostron/kubernetes-dependency-watches/client"
+       extensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+       extensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
        "k8s.io/apimachinery/pkg/api/equality"
        "k8s.io/apimachinery/pkg/api/errors"
        "k8s.io/apimachinery/pkg/api/meta"
        metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
        "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+       "k8s.io/apimachinery/pkg/labels"
        "k8s.io/apimachinery/pkg/runtime"
        "k8s.io/apimachinery/pkg/runtime/schema"
        "k8s.io/apimachinery/pkg/types"
@@ -36,8 +39,9 @@ import (
 )
 
 const (
-       ControllerName string = "policy-template-sync"
-       policyFmtStr   string = "policy: %s/%s"
+       ControllerName  string = "policy-template-sync"
+       policyFmtStr    string = "policy: %s/%s"
+       PolicyTypeLabel string = "policy.open-cluster-management.io/policy-type"
 )
 
 var log = ctrl.Log.WithName(ControllerName)
@@ -551,30 +555,36 @@ func (r *PolicyReconciler) cleanUpExcessTemplates(
        parentTemplates []string,
 ) error {
        // build list of policy template resources to check
-       tmplCRDs, err := dClient.Resource(schema.GroupVersionResource{
-               Group:    "apiextensions.k8s.io",
-               Version:  "v1",
-               Resource: "customresourcedefinitions",
-       }).List(ctx, metav1.ListOptions{
-               LabelSelector: "policy.open-cluster-management.io/policy-type=template",
-       })
-       if err != nil {
-               return fmt.Errorf("error getting policy template CRDs: %w", err)
-       }
-
+       crdLabelSelector := labels.SelectorFromSet(map[string]string{PolicyTypeLabel: "template"})
+       crdsv1 := extensionsv1.CustomResourceDefinitionList{}
        tmplGVRs := []schema.GroupVersionResource{}
 
-       for _, crd := range tmplCRDs.Items {
-               versions, ok := crd.Object["spec"].(map[string]interface{})["versions"].([]interface{})
-               if !ok {
-                       return fmt.Errorf("template CRD is formatted incorrectly: no versions found")
+       err := r.List(ctx, &crdsv1, &client.ListOptions{LabelSelector: crdLabelSelector})
+       if err == nil {
+               for _, crd := range crdsv1.Items {
+                       tmplGVRs = append(tmplGVRs, schema.GroupVersionResource{
+                               Group:    crd.Spec.Group,
+                               Resource: crd.Spec.Names.Plural,
+                               // TODO: Guard against empty versions. Not sure if that's even possible.
+                               Version: crd.Spec.Versions[0].Name,
+                       })
+               }
+       } else {
+               // TODO Detect if the error is because v1 CRDs aren't supported on this cluster.
+               // If so, try the v1beta1
+               crdsv1beta1 := extensionsv1beta1.CustomResourceDefinitionList{}
+
+               err := r.List(ctx, &crdsv1beta1, &client.ListOptions{LabelSelector: crdLabelSelector})
+               if err == nil {
+                       return err
                }
 
-               for _, version := range versions {
+               for _, crd := range crdsv1beta1.Items {
                        tmplGVRs = append(tmplGVRs, schema.GroupVersionResource{
-                               Group:    strings.SplitN(crd.GetName(), ".", 2)[1],
-                               Resource: strings.SplitN(crd.GetName(), ".", 2)[0],
-                               Version:  version.(map[string]interface{})["name"].(string),
+                               Group:    crd.Spec.Group,
+                               Resource: crd.Spec.Names.Plural,
+                               // TODO: Guard against empty versions. Not sure if that's even possible.
+                               Version: crd.Spec.Versions[0].Name,
                        })
                }
        }
diff --git a/main.go b/main.go
index ff27b15..0106603 100644
--- a/main.go
+++ b/main.go
@@ -23,7 +23,10 @@ import (
 
        // to ensure that exec-entrypoint and run can make use of them.
        v1 "k8s.io/api/core/v1"
+       extensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+       extensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
        "k8s.io/apimachinery/pkg/fields"
+       "k8s.io/apimachinery/pkg/labels"
        k8sruntime "k8s.io/apimachinery/pkg/runtime"
        utilruntime "k8s.io/apimachinery/pkg/util/runtime"
        "k8s.io/client-go/kubernetes"
@@ -330,10 +333,23 @@ func getManager(
        )
 
        hubRecorder := eventBroadcaster.NewRecorder(eventsScheme, v1.EventSource{Component: statussync.ControllerName})
+       crdLabelSelector := labels.SelectorFromSet(map[string]string{templatesync.PolicyTypeLabel: "template"})
 
        options.LeaderElectionID = "governance-policy-framework-addon.open-cluster-management.io"
        options.HealthProbeBindAddress = healthAddr
        options.Namespace = tool.Options.ClusterNamespace
+       options.NewCache = cache.BuilderWithOptions(
+               cache.Options{
+                       SelectorsByObject: cache.SelectorsByObject{
+                               &extensionsv1.CustomResourceDefinition{}: {
+                                       Label: crdLabelSelector,
+                               },
+                               &extensionsv1beta1.CustomResourceDefinition{}: {
+                                       Label: crdLabelSelector,
+                               },
+                       },
+               },
+       )
 
        mgr, err := ctrl.NewManager(managedCfg, options)
        if err != nil {
```

controllers/templatesync/template_sync.go
```diff
@@ -523,11 +531,89 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
+func (r *PolicyReconciler) cleanUpExcessTemplates(
+	ctx context.Context,
+	dClient dynamic.Interface,
+	instance policiesv1.Policy,
+	parentTemplates []string,
+) error {
+	// build list of policy template resources to check
+	tmplCRDs, err := dClient.Resource(schema.GroupVersionResource{
+		Group:    "apiextensions.k8s.io",
+		Version:  "v1",
```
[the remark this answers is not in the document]
willkutler: fixed with the code to use the reconciler client

willkutler:
@mprahl @JustinKuli review fixes are done!

controllers/templatesync/template_sync.go
```diff
@@ -523,11 +535,97 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
+		for _, crd := range crdsv1.Items {
+			if len(crd.Spec.Versions) > 0 {
+				tmplGVRs = append(tmplGVRs, schema.GroupVersionResource{
+					Group:    crd.Spec.Group,
+					Resource: crd.Spec.Names.Plural,
+					Version:  crd.Spec.Versions[0].Name,
+				})
+			}
+		}
+	} else if err.Error() == "no matches for kind \"CustomResourceDefinition\" in version \"apiextensions.k8s.io/v1\"" {
```
mprahl, line 573: Does this work instead?
```suggestion
	} else if err.Error() == `no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1"` {
```

mprahl, line 573: Do you know if there is anything from `k8s.io/apimachinery/pkg/api/errors` that we'd be able to use instead of string matching? I'm just worried that this is a bit brittle.

mprahl (changes requested):
Just some minor comments, but this looks great!

controllers/templatesync/template_sync.go
willkutler: I looked on friday and wasn't able to find a function in `k8s.io/apimachinery/pkg/api/errors` for this scenario. If you're worried about it I can change that block back to an `else` and have the error returned not be version-specific.

mprahl: @JustinKuli what do you think?

JustinKuli: I found https://pkg.go.dev/k8s.io/apimachinery@v0.26.1/pkg/api/meta#IsNoMatchError, it's slightly more general than what this checks for (which could be good). It's hard to tell if the string matching could break in the future.

willkutler: nice find @JustinKuli ! I didn't think to check the meta library. I'll add that instead of the string match

willkutler:
@mprahl @JustinKuli ok, fixed the constant and used the error method Justin found so I think this should be done with review changes

main.go
```diff
@@ -435,6 +442,12 @@ func getHubManager(
 				&v1.Secret{}: {
 					Field: fields.SelectorFromSet(fields.Set{"metadata.name": secretsync.SecretName}),
 				},
+				&extensionsv1.CustomResourceDefinition{}: {
```
mprahl, line 445: Sorry I didn't catch this before but this needs to be in the `getManager` method and not the `getHubManager` method. See my git diff I previously posted.

mprahl (changes requested):

mprahl:
@willkutler the change looks good. Could you please fix the linting error?

mprahl (approved):

mprahl:
@JustinKuli did you want to review this again? If not, @willkutler you'll need to dismiss his review so that PR can be merged.

JustinKuli (approved):

PR #30, opened 2023-02-20
Recover owner references that are removed from policy templates

willkutler:
Previously if the owner references array was removed from the metadata of a policy template, there was no way to recoveer them. With recent commits this no longer caused an index out of bounds panic, but we still want to keep the owner references, so this PR adds code to restore them if they are removed.

refs: https://issues.redhat.com/browse/ACM-3027

### DIFF
--- a/test/e2e/case16_restore_owner_refs_test.go
+++ b/test/e2e/case16_restore_owner_refs_test.go
@@ -1,0 +1,58 @@
+// Copyright (c) 2020 Red Hat, Inc.
+// Copyright Contributors to the Open Cluster Management project
+
+package e2e
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"open-cluster-management.io/governance-policy-propagator/test/utils"
+)
+
+var _ = Describe("Test owner reference recovery", func() {
+	const (
+		case16PolicyName            string = "case16-test-policy"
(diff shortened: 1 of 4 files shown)


'
```diff
@@ -0,0 +1,36 @@
+# This is a combination of 2 commits.
```
JustinKuli, line 1: You probably accidentally committed this 😄

test/e2e/case16_restore_owner_refs_test.go
```diff
@@ -0,0 +1,62 @@
...
+		utils.ListWithTimeout(clientManagedDynamic, gvrPolicy, opt, 0, true, defaultTimeoutSeconds)
+
+		By("Deleting the config policy in ns:" + clusterNamespaceOnHub)
+		_, err = kubectlManaged("delete", "-f", case16PatchConfigPolicyYaml, "-n", clusterNamespaceOnHub,
+			"--ignore-not-found=true")
+		Expect(err).To(BeNil())
+	})
+	It("Should restore owner references that are edited out of the child config policy", func() {
+		By("Patching config policy to remove owner references")
+		_, err := kubectlManaged("apply", "-f", case16PatchConfigPolicyYaml, "-n", clusterNamespace)
```
JustinKuli, line 43: I'm worried that this `apply` doesn't actually remove the owner references. I would rather have this be a `patch` that specifically clears the owner references... something like
```
kubectly patch <> --type=merge -p={"metadata":{"ownerReferences":[]}}
```

JustinKuli (dismissed):
LGTM!

JustinKuli (approved):

PR #33, opened 2023-02-22
Implement syncing for `ConstraintTemplates`

dhaiducek:
ConstraintTemplates are a concept of the Open Policy Agent, handled by
the Gatekeeper framework. They are clusterwide so require some special
handling to sync, including manual cleanup since controller garbage
collection only cleans up owned objects in the same namespace. They
also do not have a status, so their creation/update generates an event
for the status sync to pick up.

ref: https://issues.redhat.com/browse/ACM-3327

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,6 +47,7 @@ MANAGED_CONFIG ?= $(PWD)/kubeconfig_managed
 # Set the Kind version tag
 ifeq ($(KIND_VERSION), minimum)
 	KIND_ARGS = --image kindest/node:v1.19.16
+	E2E_FILTER = --label-filter="!skip-minimum"
 else ifneq ($(KIND_VERSION), latest)
 	KIND_ARGS = --image kindest/node:$(KIND_VERSION)
 else
@@ -292,7 +293,7 @@ e2e-test: e2e-dependencies
 	$(GINKGO) -v --fail-fast --slow-spec-threshold=10s $(E2E_TEST_ARGS) test/e2e
 
 .PHONY: e2e-test-coverage
-e2e-test-coverage: E2E_TEST_ARGS = --json-report=report_e2e.json --output-dir=.
+e2e-test-coverage: E2E_TEST_ARGS = --json-report=report_e2e.json --output-dir=. $(E2E_FILTER)
 e2e-test-coverage: e2e-run-instrumented e2e-test e2e-stop-instrumented
 
 .PHONY: e2e-build-instrumented
--- a/Makefile
+++ b/Makefile
@@ -290,19 +290,24 @@ e2e-dependencies:
 
 .PHONY: e2e-test
 e2e-test: e2e-dependencies
-	$(GINKGO) -v --fail-fast --slow-spec-threshold=10s $(E2E_TEST_ARGS) test/e2e
+	$(GINKGO) -v --fail-fast $(E2E_TEST_ARGS) test/e2e
 
 .PHONY: e2e-test-coverage
 e2e-test-coverage: E2E_TEST_ARGS = --json-report=report_e2e.json --output-dir=. $(E2E_FILTER)
 e2e-test-coverage: e2e-run-instrumented e2e-test e2e-stop-instrumented
+
+.PHONY: e2e-test-coverage-foreground
+e2e-test-coverage-foreground: LOG_REDIRECT = 
+e2e-test-coverage-foreground: e2e-test-coverage
 
 .PHONY: e2e-build-instrumented
 e2e-build-instrumented:
 	go test -covermode=atomic -coverpkg=$(shell cat go.mod | head -1 | cut -d ' ' -f 2)/... -c -tags e2e ./ -o build/_output/bin/$(IMG)-instrumented
 
 .PHONY: e2e-run-instrumented
+LOG_REDIRECT ?= &>build/_output/controller.log
 e2e-run-instrumented: e2e-build-instrumented
-	HUB_CONFIG=$(HUB_CONFIG) MANAGED_CONFIG=$(MANAGED_CONFIG) MANAGED_CLUSTER_NAME=$(MANAGED_CLUSTER_NAME) ./build/_output/bin/$(IMG)-instrumented -test.run "^TestRunMain$$" -test.coverprofile=$(COVERAGE_E2E_OUT) &>build/_output/controller.log &
+	HUB_CONFIG=$(HUB_CONFIG) MANAGED_CONFIG=$(MANAGED_CONFIG) MANAGED_CLUSTER_NAME=$(MANAGED_CLUSTER_NAME) ./build/_output/bin/$(IMG)-instrumented -test.run "^TestRunMain$$" -test.coverprofile=$(COVERAGE_E2E_OUT) $(LOG_REDIRECT) &
 
 .PHONY: e2e-stop-instrumented
 e2e-stop-instrumented:
--- a/Makefile
+++ b/Makefile
@@ -243,11 +243,11 @@ kind-bootstrap-cluster-dev: kind-create-cluster install-crds install-resources
 kind-deploy-controller:
 	@echo installing $(IMG)
 	-kubectl create ns $(KIND_NAMESPACE) --kubeconfig=$(MANAGED_CONFIG)
-	kubectl create secret -n $(KIND_NAMESPACE) generic hub-kubeconfig --from-file=kubeconfig=$(HUB_CONFIG_INTERNAL) --kubeconfig=$(MANAGED_CONFIG)
+	-kubectl create secret -n $(KIND_NAMESPACE) generic hub-kubeconfig --from-file=kubeconfig=$(HUB_CONFIG_INTERNAL) --kubeconfig=$(MANAGED_CONFIG)
 	kubectl apply -f deploy/operator.yaml -n $(KIND_NAMESPACE) --kubeconfig=$(MANAGED_CONFIG)
 
 .PHONY: kind-deploy-controller-dev
-kind-deploy-controller-dev: kind-deploy-controller
+kind-deploy-controller-dev: kind-deploy-controller build-images
 	@echo Pushing image to KinD cluster
 	kind load docker-image $(REGISTRY)/$(IMG):$(TAG) --name $(KIND_NAME)
 	@echo "Patch deployment image"
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -10,6 +10,8 @@ import (
 	"strings"
 
 	"github.com/go-logr/logr"
+	gktemplatesv1 "github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1"
+	gktemplatesv1beta1 "github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1beta1"
 	"github.com/prometheus/client_golang/prometheus"
 	depclient "github.com/stolostron/kubernetes-dependency-watches/client"
 	extensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
@@ -207,6 +209,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 		return reconcile.Result{}, nil
 	}
 
+	// Handle dependencies that apply to the parent policy
 	allDeps := make(map[depclient.ObjectIdentifier]string)
 	topLevelDeps := make(map[depclient.ObjectIdentifier]string)
 
@@ -268,10 +271,14 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 			continue
 		}
 
+		// Special handling booleans, whether this template is:
+		// - ContraintTemplate handled by Gatekeeper
+		// - Cluster scoped
 		isGkConstraintTemplate := gvk.Group == utils.GvkConstraintTemplate.Group &&
 			gvk.Kind == utils.GvkConstraintTemplate.Kind
 		isClusterScoped := isGkConstraintTemplate || gvk.Group == utils.GConstraint
 
+		// Handle dependencies that apply to the current policy-template
 		depConflictErr := false
 
 		// use copy of dependencies scoped only to this template
@@ -372,7 +379,39 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 			continue
 		}
 
-		// reject if not configuration policy and has templates
+		// Check for whether the CRD associated with this policy template has a policy-type=template
+		// label, signaling that it should be synced
+		hasTemplateLabel, err := r.hasPolicyTemplateLabel(ctx, rsrc)
+		if err != nil {
+			reqLogger.Error(err, fmt.Sprintf("Failed to retrieve CRD %s", rsrc.GroupResource().String()))
+
+			policySystemErrorsCounter.WithLabelValues(request.Name, tName, "get-error").Inc()
+
+			// The CRD should exist since it was found in the mapping previously; Requeue this template
+			return reconcile.Result{}, err
+		}
+
+		// If no policy-type=template label AND the GroupKind is not on the explicit allow list, don't
+		// sync this template
+		if !hasTemplateLabel && !utils.IsAllowedPolicy(gvk.GroupKind()) {
+			errMsg := fmt.Sprintf("policy-template kind is not supported: %s", gvk.String())
+			err := fmt.Errorf(errMsg)
+
+			resultError = err
+
+			r.emitTemplateError(instance, tIndex, tName, isClusterScoped, errMsg)
+			tLogger.Error(err, "Unsupported policy-template kind found in object definition",
+				"group", gvk.Group,
+				"version", gvk.Version,
+				"kind", gvk.Kind,
+			)
+
+			policyUserErrorsCounter.WithLabelValues(instance.Name, tName, "crd-error").Inc()
+
+			continue
+		}
+
+		// reject if not configuration policy and has template strings
 		if gvk.Kind != "ConfigurationPolicy" {
 			// if not configuration policies, do a simple check for templates {{hub and reject
 			// only checking for hub and not {{ as they could be valid cases where they are valid chars.
@@ -420,6 +459,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 			continue
 		}
 
+		// Collect list of dependent policies
 		childTemplates = append(childTemplates, depclient.ObjectIdentifier{
 			Group:     gvk.Group,
 			Version:   gvk.Version,
@@ -458,22 +498,16 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 
 				// Handle adding metadata labels
 				labels := tObjectUnstructured.GetLabels()
-
 				if labels == nil {
-					labels = map[string]string{
-						"cluster-name":               instance.GetLabels()[common.ClusterNameLabel],
-						common.ClusterNameLabel:      instance.GetLabels()[common.ClusterNameLabel],
-						"cluster-namespace":          instance.GetLabels()[common.ClusterNamespaceLabel],
-						common.ClusterNamespaceLabel: instance.GetLabels()[common.ClusterNamespaceLabel],
-					}
-				} else {
-					labels["cluster-name"] = instance.GetLabels()[common.ClusterNameLabel]
-					labels[common.ClusterNameLabel] = instance.GetLabels()[common.ClusterNameLabel]
-					labels["cluster-namespace"] = instance.GetLabels()[common.ClusterNamespaceLabel]
-					labels[common.ClusterNamespaceLabel] = instance.GetLabels()[common.ClusterNamespaceLabel]
+					labels = map[string]string{}
 				}
 
-				// set label to identify parent policy for this template
+				labels["cluster-name"] = instance.GetLabels()[common.ClusterNameLabel]
+				labels[common.ClusterNameLabel] = instance.GetLabels()[common.ClusterNameLabel]
+				labels["cluster-namespace"] = instance.GetLabels()[common.ClusterNamespaceLabel]
+				labels[common.ClusterNamespaceLabel] = instance.GetLabels()[common.ClusterNamespaceLabel]
+
+				// Set label to identify parent policy for this template object
 				labels[utils.ParentPolicyLabel] = instance.GetName()
 
 				tObjectUnstructured.SetLabels(labels)
@@ -494,17 +528,18 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 				}
 
 				successMsg := fmt.Sprintf("Policy template %s created successfully", tName)
-				tLogger.Info("Policy template created successfully", "PolicyTemplateName", tName)
+
+				tLogger.Info("Policy template created successfully")
 
 				// Handle clusterwide objects
 				if isClusterScoped {
 					addFinalizer = true
 
-					reqLogger.V(2).Info("Finalizer required for "+gvk.Kind, "PolicyTemplateName", tName)
+					tLogger.V(2).Info("Finalizer required for " + gvk.Kind)
 
-					// The ConstraintTemplate does not generate status, so we need to generate an event for it
+					// The ConstraintTemplate does not generate status, so we need to generate an event for it.
 					if isGkConstraintTemplate {
-						tLogger.Info("Emitting status event for "+gvk.Kind, "PolicyTemplateName", tName)
+						tLogger.Info("Emitting status event for " + gvk.Kind)
 						msg := fmt.Sprintf("%s %s was created successfully", gvk.Kind, tName)
 						r.emitTemplateSuccess(instance, tIndex, tName, isClusterScoped, msg)
 					}
@@ -570,11 +605,11 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 			break // just get the first ownerReference, if there are any at all
 		}
 
-		parentPolicyLabel := eObject.GetLabels()[utils.ParentPolicyLabel]
+		parentPolicy := eObject.GetLabels()[utils.ParentPolicyLabel]
 
 		// If the owner reference has been unset but the template is still managed by this policy
 		// instance, recover the owner reference (skip this for cluster scoped objects)
-		if !isClusterScoped && refName == "" && parentPolicyLabel == instance.GetName() {
+		if !isClusterScoped && refName == "" && parentPolicy == instance.GetName() {
 			plcOwnerReferences := *metav1.NewControllerRef(instance, schema.GroupVersionKind{
 				Group:   policiesv1.SchemeGroupVersion.Group,
 				Version: policiesv1.SchemeGroupVersion.Version,
@@ -591,7 +626,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 
 		// If there's no owner reference name, set it to the current parent policy label on the object
 		if refName == "" {
-			refName = parentPolicyLabel
+			refName = parentPolicy
 		}
 
 		// Violation when object reference (or parent policy label on the object if there's no owner
@@ -624,7 +659,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 			continue
 		}
 
-		// Fill in defaults set the by ConstraintTemplate CRD to ensure the spec comparison below is correct.
+		// Fill in defaults set by the ConstraintTemplate CRD to ensure the spec comparison below is correct.
 		if isGkConstraintTemplate {
 			err := utils.ApplyObjectDefaults(*r.Scheme, tObjectUnstructured)
 			if err != nil {
@@ -672,11 +707,11 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 			if isClusterScoped {
 				addFinalizer = true
 
-				reqLogger.V(2).Info("Finalizer required for "+gvk.Kind, "PolicyTemplateName", tName)
+				reqLogger.V(2).Info("Finalizer required for " + gvk.Kind)
 
 				// The ConstraintTemplate does not generate status, so we need to generate an event for it
 				if isGkConstraintTemplate {
-					tLogger.Info("Emitting status event for "+gvk.Kind, "PolicyTemplateName", tName)
+					tLogger.Info("Emitting status event for " + gvk.Kind)
 					msg := fmt.Sprintf("%s %s was updated successfully", gvk.Kind, tName)
 					r.emitTemplateSuccess(instance, tIndex, tName, isClusterScoped, msg)
 				}
@@ -703,9 +738,9 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 			tLogger.Info("Existing object matches the policy template")
 		}
 
-		// If we got to this point, the reconcile succeeded and a finalizer would be required for
-		// existing clusterwide objects
 		if isClusterScoped {
+			// If we got to this point, the reconcile succeeded and a finalizer would be required for
+			// existing clusterwide objects
 			addFinalizer = true
 		}
 	}
@@ -779,52 +814,147 @@ func (r *PolicyReconciler) cleanUpExcessTemplates(
 	instance policiesv1.Policy,
 	templateNames []string,
 ) error {
-	crdLabelSelector := labels.SelectorFromSet(map[string]string{utils.PolicyTypeLabel: "template"})
-	crdsv1 := extensionsv1.CustomResourceDefinitionList{}
-	tmplGVRs := []schema.GroupVersionResource{}
+	var errorList utils.ErrList
 
-	// build list of GVRs for templates to check the parent label on
-	err := r.List(ctx, &crdsv1, &client.ListOptions{LabelSelector: crdLabelSelector})
+	reqLogger := log.WithValues("Request.Namespace", instance.Namespace, "Request.Name", instance.Name)
+
+	// GVR with scope specified
+	type gvrScoped struct {
+		gvr        schema.GroupVersionResource
+		namespaced bool
+	}
+
+	tmplGVRs := []gvrScoped{}
+
+	// Query for ConstraintTemplates and collect the GroupVersionResource for each Constraint
+	gkConstraintTemplateListv1 := gktemplatesv1.ConstraintTemplateList{}
+
+	err := r.List(ctx, &gkConstraintTemplateListv1, &client.ListOptions{})
+	if err == nil {
+		// Add the ConstraintTemplate to the GVR list
+		if len(gkConstraintTemplateListv1.Items) > 0 {
+			tmplGVRs = append(tmplGVRs, gvrScoped{
+				gvr: schema.GroupVersionResource{
+					Group:    utils.GvkConstraintTemplate.Group,
+					Resource: "constrainttemplates",
+					Version:  "v1",
+				},
+				namespaced: false,
+			})
+		}
+		// Iterate over the ConstraintTemplates to gather the Constraints on the cluster
+		for _, gkCT := range gkConstraintTemplateListv1.Items {
+			tmplGVRs = append(tmplGVRs, gvrScoped{
+				gvr: schema.GroupVersionResource{
+					Group:    utils.GConstraint,
+					Resource: strings.ToLower(gkCT.Spec.CRD.Spec.Names.Kind),
+					Version:  "v1beta1",
+				},
+				namespaced: false,
+			})
+		}
+	} else if meta.IsNoMatchError(err) {
+		// If there's no v1 ConstraintTemplate, try the v1beta1 version
+		gkConstraintTemplateListv1beta1 := gktemplatesv1beta1.ConstraintTemplateList{}
+		err := r.List(ctx, &gkConstraintTemplateListv1beta1, &client.ListOptions{})
+		if err == nil {
+			// Add the ConstraintTemplate to the GVR list
+			if len(gkConstraintTemplateListv1.Items) > 0 {
+				tmplGVRs = append(tmplGVRs, gvrScoped{
+					gvr: schema.GroupVersionResource{
+						Group:    utils.GvkConstraintTemplate.Group,
+						Resource: "constrainttemplates",
+						Version:  "v1beta1",
+					},
+					namespaced: false,
+				})
+			}
+			// Iterate over the ConstraintTemplates to gather the Constraints on the cluster
+			for _, gkCT := range gkConstraintTemplateListv1beta1.Items {
+				tmplGVRs = append(tmplGVRs, gvrScoped{
+					gvr: schema.GroupVersionResource{
+						Group:    utils.GConstraint,
+						Resource: strings.ToLower(gkCT.Spec.CRD.Spec.Names.Kind),
+						Version:  "v1beta1",
+					},
+					namespaced: false,
+				})
+			}
+			// Log and ignore other errors to allow cleanup to continue since Gatekeeper may not be installed
+		} else {
+			reqLogger.Info(fmt.Sprintf("Ignoring ConstraintTemplate cleanup error: %s", err.Error()))
+		}
+	} else {
+		reqLogger.Info(fmt.Sprintf("Ignoring ConstraintTemplate cleanup error: %s", err.Error()))
+	}
+
+	// Query for CRDs with policy-type label
+	crdQuery := client.ListOptions{
+		LabelSelector: labels.SelectorFromSet(map[string]string{utils.PolicyTypeLabel: "template"}),
+	}
+
+	// Build list of GVRs for objects to check the parent label on, falling back to v1beta1
+	crdsv1 := extensionsv1.CustomResourceDefinitionList{}
+
+	err = r.List(ctx, &crdsv1, &crdQuery)
 	if err == nil {
 		for _, crd := range crdsv1.Items {
 			if len(crd.Spec.Versions) > 0 {
-				tmplGVRs = append(tmplGVRs, schema.GroupVersionResource{
-					Group:    crd.Spec.Group,
-					Resource: crd.Spec.Names.Plural,
-					Version:  crd.Spec.Versions[0].Name,
+				tmplGVRs = append(tmplGVRs, gvrScoped{
+					gvr: schema.GroupVersionResource{
+						Group:    crd.Spec.Group,
+						Resource: crd.Spec.Names.Plural,
+						Version:  crd.Spec.Versions[0].Name,
+					},
+					namespaced: crd.Spec.Scope == extensionsv1.NamespaceScoped,
 				})
 			}
 		}
 	} else if meta.IsNoMatchError(err) {
 		crdsv1beta1 := extensionsv1beta1.CustomResourceDefinitionList{}
-		err := r.List(ctx, &crdsv1beta1, &client.ListOptions{LabelSelector: crdLabelSelector})
+		err := r.List(ctx, &crdsv1beta1, &crdQuery)
 		if err != nil {
-			return fmt.Errorf("error listing v1beta1 CRDs with %s label: %w", crdLabelSelector, err)
+			return fmt.Errorf("error listing v1beta1 CRDs with query %+v: %w", crdQuery, err)
 		}
 		for _, crd := range crdsv1beta1.Items {
 			if len(crd.Spec.Versions) > 0 {
-				tmplGVRs = append(tmplGVRs, schema.GroupVersionResource{
-					Group:    crd.Spec.Group,
-					Resource: crd.Spec.Names.Plural,
-					Version:  crd.Spec.Versions[0].Name,
+				tmplGVRs = append(tmplGVRs, gvrScoped{
+					gvr: schema.GroupVersionResource{
+						Group:    crd.Spec.Group,
+						Resource: crd.Spec.Names.Plural,
+						Version:  crd.Spec.Versions[0].Name,
+					},
+					namespaced: crd.Spec.Scope == extensionsv1beta1.NamespaceScoped,
 				})
 			}
 		}
 	} else {
-		return fmt.Errorf("error listing v1 CRDs with %s label: %w", crdLabelSelector, err)
+		return fmt.Errorf("error listing v1 CRDs with query %+v: %w", crdQuery, err)
 	}
 
-	for _, gvr := range tmplGVRs {
-		// iterate through all templates with parent label set to see if they match
-		children, err := dClient.Resource(gvr).Namespace(r.ClusterNamespace).List(ctx, metav1.ListOptions{
+	for _, gvrScoped := range tmplGVRs {
+		// Instantiate a dynamic client for the GVR
+		resourceNs := ""
+		if gvrScoped.namespaced {
+			resourceNs = r.ClusterNamespace
+		}
+
+		resClient := dClient.Resource(gvrScoped.gvr).Namespace(resourceNs)
+
+		// Iterate through all objects with parent label set to see if they
+		// match the templates in the policy
+		children, err := resClient.List(ctx, metav1.ListOptions{
 			LabelSelector: utils.ParentPolicyLabel + "=" + instance.GetName(),
 		})
 		if err != nil {
-			return fmt.Errorf("error listing %s objects: %w", gvr.String(), err)
+			errorList = append(errorList,
+				fmt.Errorf("error listing %s objects: %w", gvrScoped.gvr.String(), err))
+
+			continue
 		}
 
 		for _, tmpl := range children.Items {
-			// delete all templates with label that aren't still in the policy
+			// delete all templates with policy label that aren't still in the policy
 			found := false
 
 			for _, parentTmplName := range templateNames {
@@ -836,19 +966,16 @@ func (r *PolicyReconciler) cleanUpExcessTemplates(
 			}
 
 			if !found {
-				err := dClient.Resource(schema.GroupVersionResource{
-					Group:    gvr.Group,
-					Version:  gvr.Version,
-					Resource: gvr.Resource,
-				}).Namespace(r.ClusterNamespace).Delete(ctx, tmpl.GetName(), metav1.DeleteOptions{})
+				err := resClient.Delete(ctx, tmpl.GetName(), metav1.DeleteOptions{})
 				if err != nil {
-					return fmt.Errorf("error deleting %s object %s: %w", gvr.String(), tmpl.GetName(), err)
+					errorList = append(errorList,
+						fmt.Errorf("error deleting %s object %s: %w", gvrScoped.gvr.String(), tmpl.GetName(), err))
 				}
 			}
 		}
 	}
 
-	return nil
+	return errorList.Aggregate()
 }
 
 // processDependencies iterates through all dependencies of a template and returns an array of any that are not met
@@ -1080,6 +1207,31 @@ func getLatestStatusMessage(pol *policiesv1.Policy, tIndex int) string {
 	return tmplDetails.History[0].Message
 }
 
+// hasPolicyTemplateLabel queries the CRD for a given GroupVersionResource and returns whether it
+// has the policy-type=template label and an error if the CRD could not be retrieved.
+func (r *PolicyReconciler) hasPolicyTemplateLabel(
+	ctx context.Context, rsrc schema.GroupVersionResource,
+) (bool, error) {
+	crd := extensionsv1.CustomResourceDefinition{}
+	crdName := types.NamespacedName{
+		Name: rsrc.GroupResource().String(),
+	}
+
+	err := r.Get(ctx, crdName, &crd)
+	if err != nil {
+		// If it wasn't found, then it wasn't in the cache and doesn't have the label
+		if k8serrors.IsNotFound(err) {
+			return false, nil
+		}
+
+		return false, err
+	}
+
+	return crd.GetLabels()[utils.PolicyTypeLabel] == "template", nil
+}
+
+// hasClusterwideFinalizer returns a boolean for whether a policy has a clusterwide finalizer,
+// signaling that the controller needs to handle manual cleanup of clusterwide objects.
 func hasClusterwideFinalizer(pol *policiesv1.Policy) bool {
 	for _, finalizer := range pol.Finalizers {
 		if finalizer == utils.ClusterwideFinalizer {
@@ -1090,7 +1242,7 @@ func hasClusterwideFinalizer(pol *policiesv1.Policy) bool {
 	return false
 }
 
-// removeFinalizer iterates over the finalizers and removes the one specified from the policy.
+// removeFinalizer iterates over the finalizers and removes the finalizer specified from the policy.
 func removeFinalizer(pol *policiesv1.Policy, finalizer string) {
 	i := 0
 	finalizersLength := len(pol.Finalizers)
@@ -1110,7 +1262,7 @@ func removeFinalizer(pol *policiesv1.Policy, finalizer string) {
 func finalizerCleanup(
 	ctx context.Context, pol *policiesv1.Policy, rMapper meta.RESTMapper, dClient dynamic.Interface,
 ) error {
-	var errorList []error
+	var errorList utils.ErrList
 
 	for _, policyT := range pol.Spec.PolicyTemplates {
 		object, gvk, err := unstructured.UnstructuredJSONScheme.Decode(policyT.ObjectDefinition.Raw, nil, nil)
@@ -1153,10 +1305,5 @@ func finalizerCleanup(
 		}
 	}
 
-	var err error
-	for _, errorItem := range errorList {
-		err = fmt.Errorf("%s; %w", err.Error(), errorItem)
-	}
-
-	return err
+	return errorList.Aggregate()
 }
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -15,7 +15,7 @@ import (
 	extensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	extensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/api/equality"
-	"k8s.io/apimachinery/pkg/api/errors"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -36,18 +36,19 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	"open-cluster-management.io/governance-policy-framework-addon/controllers/utils"
 )
 
 const (
-	ControllerName    string = "policy-template-sync"
-	policyFmtStr      string = "policy: %s/%s"
-	PolicyTypeLabel   string = "policy.open-cluster-management.io/policy-type"
-	parentPolicyLabel string = "policy.open-cluster-management.io/policy"
+	ControllerName string = "policy-template-sync"
 )
 
 var log = ctrl.Log.WithName(ControllerName)
 
 //+kubebuilder:rbac:groups=policy.open-cluster-management.io,resources=*,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=templates.gatekeeper.sh,resources=constrainttemplates,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=constraints.gatekeeper.sh,resources=*,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=core,resources=events,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=list;watch
 
@@ -97,10 +98,10 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 
 	err := r.Get(ctx, request.NamespacedName, instance)
 	if err != nil {
-		if errors.IsNotFound(err) {
+		if k8serrors.IsNotFound(err) {
 			// Request object not found, could have been deleted after reconcile request.
-			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
-			// Return and don't requeue
+			// Owned namespaced objects are automatically garbage collected. Additional cleanup logic uses
+			// finalizers.
 			reqLogger.Info("Policy not found, may have been deleted, reconciliation completed")
 
 			_ = policyUserErrorsCounter.DeletePartialMatch(prometheus.Labels{"policy": request.Name})
@@ -111,6 +112,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 				reqLogger.Error(err, "Error updating dependency watcher. Ignoring the failure.")
 			}
 
+			// Return and don't requeue
 			return reconcile.Result{}, nil
 		}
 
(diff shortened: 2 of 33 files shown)


controllers/templatesync/template_sync.go
```diff
@@ -273,13 +286,27 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
 		if mapping != nil {
 			rsrc = mapping.Resource
 		} else {
 			resultError = err
-			errMsg := fmt.Sprintf("Mapping not found, please check if you have CRD deployed: %s", err)
+			errMsg := "Mapping not found, "
+			if isGatekeeper {
+				errMsg += "check if Gatekeeper is installed"
+				if gvk.Group == utils.GConstraint {
+					errMsg += " and the required " + utils.GvkConstraintTemplate.Kind + " has been deployed"
```
mprahl, line 304:
```suggestion
					errMsg += " and the required ContraintTemplate for " + utils.GvkConstraintTemplate.Kind + " has been deployed"
```

mprahl:
@dhaiducek could you please update `overrideRemediationAction` to handle the desired behavior? For example:
```go
func overrideRemediationAction(instance *policiesv1.Policy, tObjectUnstructured *unstructured.Unstructured) {
	// override RemediationAction only when it is set on parent
	if instance.Spec.RemediationAction == "" {
		return
	}

	if tObjectUnstructured.GroupVersionKind().Group == gkconstraints.Group {
		var enforcementAction string

		switch strings.ToLower(string(instance.Spec.RemediationAction)) {
		case strings.ToLower(string(policiesv1.Inform)):
			enforcementAction = "warn"
		case strings.ToLower(string(policiesv1.Enforce)):
			enforcementAction = "deny"
		default:
			return
		}

		if spec, ok := tObjectUnstructured.Object["spec"]; ok {
			specObject, ok := spec.(map[string]interface{})
			if ok {
				specObject["enforcementAction"] = enforcementAction
			}
		}
	} else if tObjectUnstructured.GroupVersionKind().Group == gktemplates.SchemeGroupVersion.Group {
		// Don't override anything if it's a ConstraintTemplate
		return
	} else {
		if spec, ok := tObjectUnstructured.Object["spec"]; ok {
			specObject, ok := spec.(map[string]interface{})
			if ok {
				specObject["remediationAction"] = string(instance.Spec.RemediationAction)
			}
		}
	}
}
```

controllers/templatesync/template_sync.go
```diff
@@ -312,8 +351,41 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
 
-		// fetch resource
-		res := dClient.Resource(rsrc).Namespace(instance.GetNamespace())
+		// Instantiate a dynamic client -- if it's a clusterwide resource, then leave off the namespace
+		var res dynamic.ResourceInterface
+		var isClusterwide bool
+
+		if isGatekeeper {
+			res = dClient.Resource(rsrc)
+			isClusterwide = true
```
mprahl, line 360: Would you be able to use `mapping` to determine if it's cluster scoped?

controllers/templatesync/template_sync.go
```diff
@@ -290,12 +317,24 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 
 			policyUserErrorsCounter.WithLabelValues(instance.Name, tName, "crd-error").Inc()
 
+			if isGkConstraintTemplate {
```
mprahl, line 320: What is there is more than one constraint or ConstraintTemplate in the policy-templates array?

controllers/templatesync/template_sync.go
```diff
@@ -110,6 +113,16 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
 	}
 
+	// Set boolean to handle finalizer
+	var needsFinalizerCleanup bool
+	var finalizerCleanupFailed bool
+
+	if instance.DeletionTimestamp != nil {
+		reqLogger.Info("Policy marked for deletion--proceeding with finalizer cleanup")
+
+		needsFinalizerCleanup = true
```
mprahl, line 123: When the parent policy is being deleted, then I think this reconcile should *only* delete the cluster-scoped policy templates and then remove the finalizer. It's currently doing a lot of work that isn't necessary since the policy templates will just be deleted after the reconcile is over.

controllers/templatesync/template_sync.go
dhaiducek: This assumes that if the ConstraintTemplate CRD isn't there, then there's no need for cleanup because Kubernetes ought to have taken care of that when the CRD disappeared? This is only being applied to ConstraintTemplates, not constraints.

mprahl, line 320: This should also be true for constraints as well.

mprahl: @dhaiducek yes, you are right. Could you please add that as a comment?

dhaiducek: > This should also be true for constraints as well.

(This is probably the 10th draft of this as I think about all of the flows 😂 ) There could be a weird scenario where a bunch of different constraints kinds are combined into a single policy. In that case, some CRDs might exist and others may not. We wouldn't want finalizer flapping in that case. For that, I might have to have a boolean on whether the finalizer is applied after going through the policy-templates...

controllers/templatesync/template_sync.go
dhaiducek: Yeah, I guess the kind isn't going to change much. 😆

controllers/templatesync/template_sync.go
dhaiducek: Good point! I'll adjust it.

controllers/templatesync/template_sync.go
dhaiducek: Holy moly! I didn't know `Scope` was in there! Okay!

controllers/templatesync/template_sync.go
```diff
@@ -389,6 +462,32 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
+
+					// The ConstraintTemplate does not generate status, so we need to generate an event for it
+					if isGkConstraintTemplate {
+						tLogger.Info("Emitting status event for "+gvk.Kind, "PolicyTemplateName", tName)
+						msg := fmt.Sprintf("%s %s was created successfully", gvk.Kind, tName)
+						r.emitTemplateSuccess(instance, tIndex, tName, msg)
+					}
+				}
+
 				err = r.handleSyncSuccess(ctx, instance, tIndex, tName, successMsg, res)
```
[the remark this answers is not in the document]
dhaiducek: I was just noticing that error--thanks!

Makefile
```diff
@@ -289,7 +289,7 @@ e2e-dependencies:
 
 .PHONY: e2e-test
 e2e-test: e2e-dependencies
-	$(GINKGO) -v --fail-fast --slow-spec-threshold=10s $(E2E_TEST_ARGS) test/e2e
```
JustinKuli, line 292: I didn't realize this was deprecated! I'll keep an eye out.

controllers/utils/utils.go
```diff
@@ -0,0 +1,22 @@
+package utils
+
+import (
+	"k8s.io/apimachinery/pkg/api/equality"
+	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+)
+
+// equivalentReplicatedPolicies compares replicated policies. Returns true if they match.
```
JustinKuli, line 24: Nit:
```suggestion
// EquivalentReplicatedPolicies compares replicated policies. Returns true if they match.
```

dhaiducek:
@mprahl Something I've been considering: Should we skip cleanup for objects the controller didn't create? I'm leaning toward "no" and cleaning up everything, but I thought I'd bring it up.

mprahl:
> @mprahl Something I've been considering: Should we skip cleanup for objects the controller didn't create? I'm leaning toward "no" and cleaning up everything, but I thought I'd bring it up.

Are you referring to a scenario where a user creates the ConstraintTemplate and constraint outside of OCM policies and then later adds it in the policy?

I think in that case, we would still delete it since the user has opted in to having an OCM policy manage those.

controllers/templatesync/template_sync.go
dhaiducek: I've created a cleanup flow and function that gets called before any template reconciles.

dhaiducek:
I've addressed comments that have been left so far. Since I need to head out, I blindly copied `overrideRemediationAction()` (thanks, @mprahl! 🎉 ), but it did override the remediation action and I updated my E2E test to pass.

The overall E2E will probably still fail since I hadn't looked into that and it looks like there might be a limitation on the dependency watcher I need to look into here also when I return:
```
E0223 13:01:54.511009   92160 client.go:437] Could not start a watch request for templates.gatekeeper.sh/v1, Kind=ConstraintTemplate, Namespace=managed, Name=case16k8srequiredlabels, error: the server could not find the requested resource
2023-02-23T13:01:54.511-0500	error	policy-template-sync	templatesync/template_sync.go:700	Error updating dependency watcher	{"Request.Namespace": "managed", "Request.Name": "case16-gk-policy", "error": "the server could not find the requested resource"}
open-cluster-management.io/governance-policy-framework-addon/controllers/templatesync.(*PolicyReconciler).Reconcile
	/Users/dhaiduce/github/open-cluster-management/grc/controllers/governance-policy-framework-addon/controllers/templatesync/template_sync.go:700
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
	/Users/dhaiduce/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.2/pkg/internal/controller/controller.go:114
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/Users/dhaiduce/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.2/pkg/internal/controller/controller.go:311
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/Users/dhaiduce/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.2/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	/Users/dhaiduce/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.2/pkg/internal/controller/controller.go:227
```

controllers/templatesync/template_sync.go
```diff
@@ -296,13 +346,27 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
 		if mapping != nil {
 			rsrc = mapping.Resource
 		} else {
 			resultError = err
-			errMsg := fmt.Sprintf("Mapping not found, please check if you have CRD deployed: %s", err)
+			errMsg := "Mapping not found, "
+			if isGatekeeper {
+				errMsg += "check if Gatekeeper is installed"
+				if gvk.Group == utils.GConstraint {
+					errMsg += " and the required ConstraintTemplate has been deployed"
```
mprahl, line 364: If the Policy has the ConstraintTemplate and the constraint, there is going to be a delay before Gatekeeper creates the CRD based on the ConstraintTemplate. My preference is to detect if a ConstraintTemplate was created previously in the loop and adjust the message.

controllers/templatesync/template_sync.go
```diff
@@ -389,7 +462,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 				}
 
 				// set label to identify parent policy for this template
-				labels[parentPolicyLabel] = instance.GetName()
+				labels[utils.ParentPolicyLabel] = instance.GetName()
 
 				tObjectUnstructured.SetLabels(labels)
 				tObjectUnstructured.SetOwnerReferences([]metav1.OwnerReference{plcOwnerReferences})
```
mprahl, line 468: Owner references shouldn't be set on the cluster-scoped policy templates.

controllers/templatesync/template_sync.go
```diff
@@ -866,3 +1043,67 @@ func getLatestStatusMessage(pol *policiesv1.Policy, tIndex int) string {
...
+	}
+
+	return false
+}
+
+// removeFinalizer iterates over the finalizers and removes the one specified from the policy.
+func removeFinalizer(pol *policiesv1.Policy, finalizer string) {
+	for idx, fzr := range pol.Finalizers {
+		if fzr == finalizer {
+			pol.Finalizers = append(pol.Finalizers[:idx], pol.Finalizers[idx+1:]...)
```
mprahl, line 1061: You either need to break after this or handle this differently since you are ranging over a slice that you are modifying in the range, which isn't safe.

mprahl:
Since @dhaiducek is on PTO, I'll rebase my local changes on this and address the feedback I left and make a subsequent PR.

mprahl:
Here are untested changes that I think should go in this PR:
```diff
From 252f2580080c43dc499888f58383f1cbcc71c9c6 Mon Sep 17 00:00:00 2001
From: mprahl <mprahl@users.noreply.github.com>
Date: Mon, 27 Feb 2023 11:28:06 -0500
Subject: [PATCH] TEMP: Dale's PR fixes

Signed-off-by: mprahl <mprahl@users.noreply.github.com>
---
 controllers/statussync/policy_status_sync.go |   2 +-
 controllers/templatesync/template_sync.go    | 156 ++++++++++++-------
 2 files changed, 98 insertions(+), 60 deletions(-)

diff --git a/controllers/statussync/policy_status_sync.go b/controllers/statussync/policy_status_sync.go
index d427e91..80ee3e5 100644
--- a/controllers/statussync/policy_status_sync.go
+++ b/controllers/statussync/policy_status_sync.go
@@ -175,7 +175,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 	// filter events to current policy instance and build map
 	eventForPolicyMap := make(map[string]*[]historyEvent)
 	// panic if regexp invalid
-	rgx := regexp.MustCompile(`(?i)^policy:\s*([A-Za-z0-9.-]+)\s*\/([A-Za-z0-9.-]+)`)
+	rgx := regexp.MustCompile(`(?i)^policy:\s*(?:([a-z0-9.-]+)/)?([a-z0-9.-]+)`)
 	for _, event := range eventList.Items {
 		// sample event.Reason -- reason: 'policy: calamari/policy-grc-rbactest-example'
 		reason := rgx.FindString(event.Reason)
diff --git a/controllers/templatesync/template_sync.go b/controllers/templatesync/template_sync.go
index 24dd274..4371d21 100644
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -41,8 +41,9 @@ import (
 )
 
 const (
-	ControllerName string = "policy-template-sync"
-	policyFmtStr   string = "policy: %s/%s"
+	ControllerName            string = "policy-template-sync"
+	policyFmtStr              string = "policy: %s/%s"
+	policyClusterScopedFmtStr string = "policy: %s"
 )
 
 var log = ctrl.Log.WithName(ControllerName)
@@ -253,6 +254,24 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 	// PolicyTemplates is not empty
 	// loop through policy templates
 	for tIndex, policyT := range instance.Spec.PolicyTemplates {
+		object, gvk, err := unstructured.UnstructuredJSONScheme.Decode(policyT.ObjectDefinition.Raw, nil, nil)
+		if err != nil {
+			resultError = err
+			errMsg := fmt.Sprintf("Failed to decode policy template with err: %s", err)
+
+			r.emitTemplateError(instance, tIndex, false, fmt.Sprintf("[template %v]", tIndex), errMsg)
+			reqLogger.Error(resultError, "Failed to decode the policy template", "templateIndex", tIndex)
+
+			policyUserErrorsCounter.WithLabelValues(instance.Name, "", "format-error").Inc()
+
+			continue
+		}
+
+		isGkConstraintTemplate := gvk.Group == utils.GvkConstraintTemplate.Group &&
+			gvk.Kind == utils.GvkConstraintTemplate.Kind
+		isClusterScoped := (gvk.Group == utils.GvkConstraintTemplate.Group &&
+			gvk.Kind == utils.GvkConstraintTemplate.Kind) || gvk.Group == utils.GConstraint
+
 		depConflictErr := false
 
 		// use copy of dependencies scoped only to this template
@@ -283,7 +302,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 				resultError = fmt.Errorf("dependency on %s has conflicting compliance states", dep.Name)
 				errMsg := fmt.Sprintf("Failed to decode policy template with err: %s", resultError)
 
-				r.emitTemplateError(instance, tIndex, fmt.Sprintf("[template %v]", tIndex), errMsg)
+				r.emitTemplateError(instance, tIndex, isClusterScoped, fmt.Sprintf("[template %v]", tIndex), errMsg)
 				reqLogger.Error(resultError, "Failed to decode the policy template", "templateIndex", tIndex)
 
 				depConflictErr = true
@@ -302,19 +321,6 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 			continue
 		}
 
-		object, gvk, err := unstructured.UnstructuredJSONScheme.Decode(policyT.ObjectDefinition.Raw, nil, nil)
-		if err != nil {
-			resultError = err
-			errMsg := fmt.Sprintf("Failed to decode policy template with err: %s", err)
-
-			r.emitTemplateError(instance, tIndex, fmt.Sprintf("[template %v]", tIndex), errMsg)
-			reqLogger.Error(resultError, "Failed to decode the policy template", "templateIndex", tIndex)
-
-			policyUserErrorsCounter.WithLabelValues(instance.Name, "", "format-error").Inc()
-
-			continue
-		}
-
 		var tName string
 		if tMetaObj, ok := object.(metav1.Object); ok {
 			tName = tMetaObj.GetName()
@@ -324,7 +330,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 			errMsg := fmt.Sprintf("Failed to get name from policy template at index %v", tIndex)
 			resultError = errors.NewBadRequest(errMsg)
 
-			r.emitTemplateError(instance, tIndex, fmt.Sprintf("[template %v]", tIndex), errMsg)
+			r.emitTemplateError(instance, tIndex, isClusterScoped, fmt.Sprintf("[template %v]", tIndex), errMsg)
 			reqLogger.Error(resultError, "Failed to process the policy template", "templateIndex", tIndex)
 
 			policyUserErrorsCounter.WithLabelValues(instance.Name, "", "format-error").Inc()
@@ -346,11 +352,6 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 
 		var rsrc schema.GroupVersionResource
 
-		// Detect ConstraintTemplate
-		isGkConstraintTemplate := gvk.Group == utils.GvkConstraintTemplate.Group &&
-			gvk.Kind == utils.GvkConstraintTemplate.Kind
-		isGatekeeper := isGkConstraintTemplate || gvk.Group == utils.GConstraint
-
 		mapping, err := rMapper.RESTMapping(gvk.GroupKind(), gvk.Version)
 
 		if mapping != nil {
@@ -358,17 +359,16 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 		} else {
 			resultError = err
 			errMsg := "Mapping not found, "
-			if isGatekeeper {
+			if isGkConstraintTemplate {
 				errMsg += "check if Gatekeeper is installed"
-				if gvk.Group == utils.GConstraint {
-					errMsg += " and the required ConstraintTemplate has been deployed"
-				}
+			} else if gvk.Group == utils.GConstraint {
+				errMsg += "check if the required ConstraintTemplate has been deployed"
 			} else {
 				errMsg += "check if you have the CRD deployed"
 			}
 			errMsg += fmt.Sprintf(": %s", err)
 
-			r.emitTemplateError(instance, tIndex, tName, errMsg)
+			r.emitTemplateError(instance, tIndex, isClusterScoped, tName, errMsg)
 			tLogger.Error(err, "Could not find an API mapping for the object definition",
 				"group", gvk.Group,
 				"version", gvk.Version,
@@ -388,7 +388,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 				errMsg := fmt.Sprintf("Templates are not supported for kind : %s", gvk.Kind)
 				resultError = errors.NewBadRequest(errMsg)
 
-				r.emitTemplateError(instance, tIndex, tName, errMsg)
+				r.emitTemplateError(instance, tIndex, isClusterScoped, tName, errMsg)
 				tLogger.Error(resultError, "Failed to process the policy template")
 
 				policyUserErrorsCounter.WithLabelValues(instance.Name, tName, "format-error").Inc()
@@ -402,8 +402,10 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 		// Instantiate a dynamic client -- if it's a clusterwide resource, then leave off the namespace
 		var res dynamic.ResourceInterface
 
-		isClusterwide := mapping.Scope == meta.RESTScopeRoot
-		if isClusterwide {
+		// Now that there is a mapping, there is a definitive answer of if it's cluster scoped rather than the
+		// previous educated guess.
+		isClusterScoped = mapping.Scope == meta.RESTScopeRoot
+		if isClusterScoped {
 			res = dClient.Resource(rsrc)
 		} else {
 			res = dClient.Resource(rsrc).Namespace(instance.GetNamespace())
@@ -416,7 +418,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 			resultError = err
 			errMsg := fmt.Sprintf("Failed to unmarshal the policy template: %s", err)
 
-			r.emitTemplateError(instance, tIndex, tName, errMsg)
+			r.emitTemplateError(instance, tIndex, isClusterScoped, tName, errMsg)
 			tLogger.Error(resultError, "Failed to unmarshal the policy template")
 
 			policySystemErrorsCounter.WithLabelValues(instance.Name, tName, "unmarshal-error").Inc()
@@ -429,7 +431,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 		if err != nil {
 			if len(dependencyFailures) > 0 {
 				// template must be pending, do not create it
-				r.emitTemplatePending(instance, tIndex, tName, generatePendingMsg(dependencyFailures))
+				r.emitTemplatePending(instance, tIndex, isClusterScoped, tName, generatePendingMsg(dependencyFailures))
 				tLogger.Info("Dependencies were not satisfied for the policy template",
 					"namespace", instance.GetNamespace(),
 					"kind", gvk.Kind,
@@ -440,11 +442,6 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 
 			if errors.IsNotFound(err) {
 				// not found should create it
-				plcOwnerReferences := *metav1.NewControllerRef(instance, schema.GroupVersionKind{
-					Group:   policiesv1.SchemeGroupVersion.Group,
-					Version: policiesv1.SchemeGroupVersion.Version,
-					Kind:    policiesv1.Kind,
-				})
 				labels := tObjectUnstructured.GetLabels()
 
 				if labels == nil {
@@ -465,7 +462,15 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 				labels[utils.ParentPolicyLabel] = instance.GetName()
 
 				tObjectUnstructured.SetLabels(labels)
-				tObjectUnstructured.SetOwnerReferences([]metav1.OwnerReference{plcOwnerReferences})
+
+				if !isClusterScoped {
+					plcOwnerReferences := *metav1.NewControllerRef(instance, schema.GroupVersionKind{
+						Group:   policiesv1.SchemeGroupVersion.Group,
+						Version: policiesv1.SchemeGroupVersion.Version,
+						Kind:    policiesv1.Kind,
+					})
+					tObjectUnstructured.SetOwnerReferences([]metav1.OwnerReference{plcOwnerReferences})
+				}
 
 				overrideRemediationAction(instance, tObjectUnstructured)
 
@@ -474,7 +479,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 					resultError = err
 					errMsg := fmt.Sprintf("Failed to create policy template: %s", err)
 
-					r.emitTemplateError(instance, tIndex, tName, errMsg)
+					r.emitTemplateError(instance, tIndex, isClusterScoped, tName, errMsg)
 					tLogger.Error(resultError, "Failed to create policy template")
 
 					policySystemErrorsCounter.WithLabelValues(instance.Name, tName, "create-error").Inc()
@@ -486,7 +491,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 				tLogger.Info("Policy template created successfully", "PolicyTemplateName", tName)
 
 				// Handle clusterwide objects
-				if isClusterwide {
+				if isClusterScoped {
 					addFinalizer = true
 
 					reqLogger.V(2).Info("Finalizer required for "+gvk.Kind, "PolicyTemplateName", tName)
@@ -495,7 +500,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 					if isGkConstraintTemplate {
 						tLogger.Info("Emitting status event for "+gvk.Kind, "PolicyTemplateName", tName)
 						msg := fmt.Sprintf("%s %s was created successfully", gvk.Kind, tName)
-						r.emitTemplateSuccess(instance, tIndex, tName, msg)
+						r.emitTemplateSuccess(instance, tIndex, isClusterScoped, tName, msg)
 					}
 				}
 
@@ -514,7 +519,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 				resultError = err
 				errMsg := fmt.Sprintf("Failed to get the object in the policy template: %s", err)
 
-				r.emitTemplateError(instance, tIndex, tName, errMsg)
+				r.emitTemplateError(instance, tIndex, isClusterScoped, tName, errMsg)
 				tLogger.Error(err, "Failed to get the object in the policy template",
 					"namespace", instance.GetNamespace(),
 					"kind", gvk.Kind,
@@ -528,7 +533,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 
 		if len(dependencyFailures) > 0 {
 			// template must be pending, need to delete it and error
-			r.emitTemplatePending(instance, tIndex, tName, generatePendingMsg(dependencyFailures))
+			r.emitTemplatePending(instance, tIndex, isClusterScoped, tName, generatePendingMsg(dependencyFailures))
 			tLogger.Info("Dependencies were not satisfied for the policy template",
 				"namespace", instance.GetNamespace(),
 				"kind", gvk.Kind,
@@ -556,9 +561,9 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 			break // just get the first ownerReference, if there are any at all
 		}
 
-		parentPolicyLabel, ok := eObject.GetLabels()["policy.open-cluster-management.io/policy"]
+		parentPolicyLabel := eObject.GetLabels()["policy.open-cluster-management.io/policy"]
 
-		if refName == "" && ok && parentPolicyLabel == instance.GetName() {
+		if !isClusterScoped && refName == "" && parentPolicyLabel == instance.GetName() {
 			// if owner ref has been unset but the template is still managed by this policy instance,
 			// recover the owner reference
 			plcOwnerReferences := *metav1.NewControllerRef(instance, schema.GroupVersionKind{
@@ -596,7 +601,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 
 			resultError = errors.NewBadRequest(errMsg)
 
-			r.emitTemplateError(instance, tIndex, tName, errMsg)
+			r.emitTemplateError(instance, tIndex, isClusterScoped, tName, errMsg)
 			tLogger.Error(resultError, "Failed to create the policy template")
 
 			policyUserErrorsCounter.WithLabelValues(instance.Name, tName, "format-error").Inc()
@@ -630,7 +635,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 				resultError = err
 				errMsg := fmt.Sprintf("Failed to update policy template %s: %s", tName, err)
 
-				r.emitTemplateError(instance, tIndex, tName, errMsg)
+				r.emitTemplateError(instance, tIndex, isClusterScoped, tName, errMsg)
 				tLogger.Error(err, "Failed to update the policy template")
 
 				policySystemErrorsCounter.WithLabelValues(instance.Name, tName, "patch-error").Inc()
@@ -641,7 +646,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 			successMsg := fmt.Sprintf("Policy template %s was updated successfully", tName)
 
 			// Handle clusterwide objects
-			if isClusterwide {
+			if isClusterScoped {
 				addFinalizer = true
 
 				reqLogger.V(2).Info("Finalizer required for "+gvk.Kind, "PolicyTemplateName", tName)
@@ -650,7 +655,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 				if isGkConstraintTemplate {
 					tLogger.Info("Emitting status event for "+gvk.Kind, "PolicyTemplateName", tName)
 					msg := fmt.Sprintf("%s %s was updated successfully", gvk.Kind, tName)
-					r.emitTemplateSuccess(instance, tIndex, tName, msg)
+					r.emitTemplateSuccess(instance, tIndex, isClusterScoped, tName, msg)
 				}
 			}
 
@@ -677,7 +682,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 
 		// If we got to this point, the reconcile succeeded and a finalizer would be required for
 		// existing clusterwide objects
-		if isClusterwide {
+		if isClusterScoped {
 			addFinalizer = true
 		}
 	}
@@ -934,14 +939,23 @@ func overrideRemediationAction(instance *policiesv1.Policy, tObjectUnstructured
 // emitTemplateSuccess performs actions that ensure correct reporting of template success in the
 // policy framework. If the policy's status already reflects the current message, then no actions
 // are taken.
-func (r *PolicyReconciler) emitTemplateSuccess(pol *policiesv1.Policy, tIndex int, tName, msg string) {
+func (r *PolicyReconciler) emitTemplateSuccess(
+	pol *policiesv1.Policy, tIndex int, clusterScoped bool, tName, msg string,
+) {
 	// check if the success is already present in the policy status - if so, return early
 	if strings.Contains(getLatestStatusMessage(pol, tIndex), msg) {
 		return
 	}
 
 	// emit the compliant event
-	policyComplianceReason := fmt.Sprintf(policyFmtStr, pol.GetNamespace(), tName)
+	var policyComplianceReason string
+
+	if clusterScoped {
+		policyComplianceReason = fmt.Sprintf(policyClusterScopedFmtStr, tName)
+	} else {
+		policyComplianceReason = fmt.Sprintf(policyFmtStr, pol.GetNamespace(), tName)
+	}
+
 	r.Recorder.Event(pol, "Normal", policyComplianceReason, "Compliant; "+msg)
 
 	// emit an informational event
@@ -951,14 +965,23 @@ func (r *PolicyReconciler) emitTemplateSuccess(pol *policiesv1.Policy, tIndex in
 // emitTemplateError performs actions that ensure correct reporting of template errors in the
 // policy framework. If the policy's status already reflects the current error, then no actions
 // are taken.
-func (r *PolicyReconciler) emitTemplateError(pol *policiesv1.Policy, tIndex int, tName, errMsg string) {
+func (r *PolicyReconciler) emitTemplateError(
+	pol *policiesv1.Policy, tIndex int, clusterScoped bool, tName, errMsg string,
+) {
 	// check if the error is already present in the policy status - if so, return early
 	if strings.Contains(getLatestStatusMessage(pol, tIndex), errMsg) {
 		return
 	}
 
 	// emit the non-compliance event
-	policyComplianceReason := fmt.Sprintf(policyFmtStr, pol.GetNamespace(), tName)
+	var policyComplianceReason string
+
+	if clusterScoped {
+		policyComplianceReason = fmt.Sprintf(policyClusterScopedFmtStr, tName)
+	} else {
+		policyComplianceReason = fmt.Sprintf(policyFmtStr, pol.GetNamespace(), tName)
+	}
+
 	r.Recorder.Event(pol, "Warning", policyComplianceReason, "NonCompliant; template-error; "+errMsg)
 
 	// emit an informational event
@@ -968,7 +991,9 @@ func (r *PolicyReconciler) emitTemplateError(pol *policiesv1.Policy, tIndex int,
 // emitTemplatePending performs actions that ensure correct reporting of pending dependencies in the
 // policy framework. If the policy's status already reflects the current status, then no actions
 // are taken.
-func (r *PolicyReconciler) emitTemplatePending(pol *policiesv1.Policy, tIndex int, tName, msg string) {
+func (r *PolicyReconciler) emitTemplatePending(
+	pol *policiesv1.Policy, tIndex int, clusterScoped bool, tName, msg string,
+) {
 	statusMsg := "Pending; " + msg
 	eventType := "Warning"
 
@@ -982,8 +1007,15 @@ func (r *PolicyReconciler) emitTemplatePending(pol *policiesv1.Policy, tIndex in
 		return
 	}
 
+	var policyComplianceReason string
+
+	if clusterScoped {
+		policyComplianceReason = fmt.Sprintf(policyClusterScopedFmtStr, tName)
+	} else {
+		policyComplianceReason = fmt.Sprintf(policyFmtStr, pol.GetNamespace(), tName)
+	}
+
 	// emit the non-compliance event
-	policyComplianceReason := fmt.Sprintf(policyFmtStr, pol.GetNamespace(), tName)
 	r.Recorder.Event(pol, eventType, policyComplianceReason, statusMsg)
 
 	// emit an informational event
@@ -1056,9 +1088,15 @@ func hasClusterwideFinalizer(pol *policiesv1.Policy) bool {
 
 // removeFinalizer iterates over the finalizers and removes the one specified from the policy.
 func removeFinalizer(pol *policiesv1.Policy, finalizer string) {
-	for idx, fzr := range pol.Finalizers {
-		if fzr == finalizer {
-			pol.Finalizers = append(pol.Finalizers[:idx], pol.Finalizers[idx+1:]...)
+	i := 0
+	finalizersLength := len(pol.Finalizers)
+
+	for i < finalizersLength {
+		if pol.Finalizers[i] == finalizer {
+			pol.Finalizers = append(pol.Finalizers[:i], pol.Finalizers[i+1:]...)
+			finalizersLength--
+		} else {
+			i++
 		}
 	}
 }
-- 
2.39.2


```

controllers/templatesync/template_sync.go
dhaiducek: Okay, I'll update it. I wasn't sure whether we wanted to leave them on for reference.

mprahl: I think they're useful but they are invalid owner references and I don't know if the Kubernetes behavior will change in the future. I think it's safer to not have them.

test/e2e/case17_gatekeeper_sync_test.go
```diff
@@ -0,0 +1,136 @@
...
+	gvrConstraint := schema.GroupVersionResource{
+		Group:    gvConstraintGroup,
+		Version:  "v1beta1",
+		Resource: caseNumber + "k8srequiredlabels",
+	}
+	BeforeAll(func() {
+		DeployGatekeeper()
+	})
+	AfterAll(func() {
+		By("Deleting policy " + policyName + " on the hub in ns:" + clusterNamespaceOnHub)
```
mprahl, line 44: Could you please also delete `case17-gk-policy-2`?

test/e2e/case17_gatekeeper_sync_test.go
```diff
@@ -0,0 +1,136 @@
...
+		_, err := kubectlHub("delete", "-f", policyYaml, "-n", clusterNamespaceOnHub)
+		var e *exec.ExitError
+		if !errors.As(err, &e) {
+			Expect(err).Should(BeNil())
+		}
```
mprahl, line 49:
```suggestion
		for _, pName := range []string{policyName, policyName2} {
			By("Deleting policy " + pName + " on the hub in ns:" + clusterNamespaceOnHub)
			err := clientHubDynamic.Resource(gvrPolicy).Namespace(clusterNamespaceOnHub).Delete(
				context.TODO(), policyName, metav1.DeleteOptions{},
			)
			if !k8serrors.IsNotFound(err) {
				Expect(err).To(BeNil())
			}

			By("Cleaning up the events for the policy " + pName)
			_, err = kubectlManaged(
				"delete",
				"events",
				"-n",
				clusterNamespace,
				"--field-selector=involvedObject.name="+pName,
				"--ignore-not-found",
			)
			Expect(err).To(BeNil())
		}
```

controllers/templatesync/template_sync.go
dhaiducek: I saw in your code update that the message is just "check if the required ConstraintTemplate has been deployed". Is that sufficient here or was there another message you were thinking about having?

I ask because I'm not sure there's a huge bump in value to adjusting the message based on whether the ConstraintTemplate is in the policy. Plus, the behavior would differ based on whether the ConstraintTemplate were before or after the Constraints, which I'm not sure is desirable. What do you think?

mprahl: Yeah, I think that is sufficient. I agree, it wasn't worth the value of adding that additional code.

controllers/templatesync/template_sync.go
```diff
@@ -866,3 +1058,73 @@ func getLatestStatusMessage(pol *policiesv1.Policy, tIndex int) string {
...
+
+// finalizerCleanup handles any steps required for cleaning up objects on the cluster prior to
+// removing the policy finalizer.
+func finalizerCleanup(
+	ctx context.Context, pol *policiesv1.Policy, rMapper meta.RESTMapper, dClient dynamic.Interface,
+) error {
+	for _, policyT := range pol.Spec.PolicyTemplates {
+		object, gvk, err := unstructured.UnstructuredJSONScheme.Decode(policyT.ObjectDefinition.Raw, nil, nil)
+		if err != nil {
+			return fmt.Errorf("failed to decode policy template with error: %w", err)
```
mprahl, line 1095: Do you think we should `continue` here to handle the case where the user modifies the policy with a syntax error and then deletes it?

If not, then I think there should be a status set on the policy indicating that there is an invalid policy template so the user can fix it.

controllers/templatesync/template_sync.go
```diff
@@ -153,6 +155,53 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
+
+		// Cleanup succeeded--remove finalizer
+		reqLogger.Info("Cleanup complete--removing clusterwide cleanup finalizer")
+		removeFinalizer(instance, utils.ClusterwideFinalizer)
+
+		err = r.Update(ctx, instance)
+		if err != nil {
+			reqLogger.Error(err, "Failed to update policy when removing finalizers")
+
+			return reconcile.Result{}, err
```
mprahl, line 200: Should this increment the system error counter?

controllers/templatesync/template_sync.go
```diff
@@ -176,7 +225,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 		}
 
 		existingDep, ok := topLevelDeps[depID]
-		if ok && existingDep != dep.Compliance {
+		if ok && existingDep != string(dep.Compliance) {
```
mprahl, line 228: What prompted this change since it's already a string?

Also, @JustinKuli as a side note, should this also be changing these to lowercase for comparison because I don't see any restrictions of the user specifying `complaint` vs `Compliant`.

controllers/templatesync/template_sync.go
```diff
@@ -488,7 +589,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 		}
 
 		// violation if object reference and policy don't match
-		if instance.GetName() != refName {
+		if !isClusterScoped && instance.GetName() != refName {
```
mprahl, line 592: We still need to catch the case for when the ConstraintTemplate was created outside of the policy or is a duplicate in general. I suggest this:
```suggestion
		if (parentPolicyLabel != "" && parentPolicyLabel != instance.GetName()) || (!isClusterScoped && instance.GetName() != refName) {
```

The code inside this if block would need to be adjusted slightly though.

controllers/templatesync/template_sync.go
dhaiducek: The `Compliance` in the new propagator was updated to an `enum`, so it has to be converted to a string to compare it.

dhaiducek: (I bumped the Propagator version to pick up updated test utils.)

controllers/templatesync/template_sync.go
```diff
@@ -828,13 +1018,15 @@ func (r *PolicyReconciler) handleSyncSuccess(
...
+	gvr *schema.GroupVersionKind,
 ) error {
 	if msg != "" {
 		r.Recorder.Event(pol, "Normal", "PolicyTemplateSync", msg)
 	}
 
-	// Only do additional steps if a template-error is the most recent status
-	if !strings.Contains(getLatestStatusMessage(pol, tIndex), "template-error;") {
+	// Only do additional steps if a template-error is the most recent status or this isn't an OCM policy
+	if gvr.Group != "policy.open-cluster-management.io" ||
```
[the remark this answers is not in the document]
dhaiducek: I actually have this one locally--not sure how it didn't get pushed yet. I'm using `common.APIGroup`, but this one might be more accurate. Do you have a preference?

controllers/templatesync/template_sync.go
dhaiducek: Good point--for cleanup, we should probably try everything rather than returning. I'll implement an aggregated error here to gather all the errors together to return.

controllers/templatesync/template_sync.go
dhaiducek: The enum applies to dependencies, so the case checking shouldn't be required: https://github.com/open-cluster-management-io/governance-policy-propagator/pull/80/files

dhaiducek: Alternatively, should `existingDep` be converted to `policiesv1.ComplianceState` instead?

dhaiducek:
Fingers crossed, the latest push should might maybe work-ish. 😬 The technical debt in this PR currently:
- (I might look into this this afternoon) Double check logic that adding/removing a Constraint[Template] policy-template from a policy is handled properly
- Handling for older Kubernetes/older Gatekeeper (right now, I have a `skip-minimum` tag that skips the test entirely); Gatekeeper v3.4 added a PodDisruptionBudget that isn't available until Kubernetes v1.21

controllers/templatesync/template_sync.go
```diff
@@ -243,50 +312,43 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
 			reqLogger.Error(resultError, "Failed to process the policy template", "templateIndex", tIndex)
 
 			policyUserErrorsCounter.WithLabelValues(instance.Name, "", "format-error").Inc()
 
 			continue
 		}
 
+		childTemplateNs := ""
+
+		if !isClusterScoped {
```
mprahl, line 343: Could perform this check after `isClusterScoped` is confirmed (i.e. after `isClusterScoped = mapping.Scope == meta.RESTScopeRoot`)?

controllers/templatesync/template_sync.go
```diff
@@ -866,3 +1072,84 @@ func getLatestStatusMessage(pol *policiesv1.Policy, tIndex int) string {
...
+			if err != nil && !k8serrors.IsNotFound(err) {
+				policySystemErrorsCounter.WithLabelValues(pol.Name, tName, "delete-error").Inc()
+
+				errorList = append(errorList, fmt.Errorf("failed to delete "+gvk.Kind+" with error: %w", err))
+			}
+		}
+	}
+
+	var err error
+	for _, errorItem := range errorList {
```
mprahl, line 1150: Optional, but this is a pretty cool way to aggregate errors:
https://pkg.go.dev/k8s.io/apimachinery@v0.23.9/pkg/util/errors#Aggregate

mprahl:
@dhaiducek to confirm, you are planning to update `cleanUpExcessTemplates` in a separate PR right?

dhaiducek:
> @dhaiducek to confirm, you are planning to update `cleanUpExcessTemplates` in a separate PR right?

Yes. I started poking at it yesterday, and it's a bit complex to get multiple CRDs with the current flows. I think it'd be best to address that in a subsequent PR, assuming this PR looks good once I address the current reviews.

mprahl:
@dhaiducek I just noticed that the template sync *always* updates the `ConstraintTemplate` on every reconcile when `spec.crd.spec.validation.legacySchema` is not set since that value is set `true` when the Kubernetes API returns the `ConstraintTemplate`.

controllers/templatesync/template_sync.go
dhaiducek: Oh nice--I saw things coming in Go 1.20, but I'll poke at this instead!

controllers/templatesync/template_sync.go
dhaiducek: Yes, then we can reuse the namespace from the dynamic client also. Good find!

controllers/templatesync/template_sync.go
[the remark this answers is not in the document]
dhaiducek: I went with `policiesv1`. Probably doesn't make much difference here, but in other places it makes more sense to use `common.APIGroup` because it's a `const` used inside of `const` strings for the label keys.

mprahl:
> @dhaiducek I just noticed that the template sync _always_ updates the `ConstraintTemplate` on every reconcile when `spec.crd.spec.validation.legacySchema` is not set since that value is set `true` when the Kubernetes API returns the `ConstraintTemplate`.

@dhaiducek this should do the trick:
```diff
diff --git a/controllers/templatesync/template_sync.go b/controllers/templatesync/template_sync.go
index d481db3..36181c4 100644
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -630,6 +630,14 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
                        continue
                }
 
+               // Fill in defaults set the by ConstraintTemplate CRD to ensure the spec comparison below is correct.
+               if isGkConstraintTemplate {
+                       err := utils.ApplyDefaults(*r.Scheme, tObjectUnstructured)
+                       if err != nil {
+                               log.Error(err, "Failed to apply defaults to the ConstraintTemplate for comparison. Continuing anyways.")
+                       }
+               }
+
                overrideRemediationAction(instance, tObjectUnstructured)
                // got object, need to compare both spec and annotation and update
                eObjectUnstructured := eObject.UnstructuredContent()
@@ -723,9 +731,15 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 
        if err != nil {
                resultError = err
-               reqLogger.Error(resultError, "Error updating dependency watcher")
 
-               policySystemErrorsCounter.WithLabelValues(instance.Name, "", "client-error").Inc()
+               if k8serrors.IsNotFound(err) {
+                       reqLogger.Error(resultError, "Error updating dependency watcher likely due to a missing CRD.")
+                       policyUserErrorsCounter.WithLabelValues(instance.Name, "", "crd-error").Inc()
+               } else {
+                       reqLogger.Error(resultError, "Error updating dependency watcher")
+                       policySystemErrorsCounter.WithLabelValues(instance.Name, "", "client-error").Inc()
+               }
+
        }
 
        err = r.cleanUpExcessTemplates(ctx, dClient, *instance, templateNames)
diff --git a/controllers/utils/utils.go b/controllers/utils/utils.go
index 1e22a6c..5b84e4f 100644
--- a/controllers/utils/utils.go
+++ b/controllers/utils/utils.go
@@ -1,7 +1,12 @@
 package utils
 
 import (
+       "encoding/json"
+       "fmt"
+
        "k8s.io/apimachinery/pkg/api/equality"
+       "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+       "k8s.io/apimachinery/pkg/runtime"
        "k8s.io/apimachinery/pkg/runtime/schema"
        policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
        "open-cluster-management.io/governance-policy-propagator/controllers/common"
@@ -35,3 +40,38 @@ func EquivalentReplicatedPolicies(plc1 *policiesv1.Policy, plc2 *policiesv1.Poli
        // Compare the specs
        return equality.Semantic.DeepEqual(plc1.Spec, plc2.Spec)
 }
+
+func ApplyDefaults(scheme runtime.Scheme, object *unstructured.Unstructured) error {
+       errDefault := "an unexpected error occurred while filling in default fields to the ConstraintTemplate: %w"
+
+       objectTyped, err := scheme.New(object.GroupVersionKind())
+       if err != nil {
+               if runtime.IsNotRegisteredError(err) {
+                       return nil
+               }
+
+               return err
+       }
+
+       err = runtime.DefaultUnstructuredConverter.FromUnstructured(object.Object, objectTyped)
+       if err != nil {
+               return fmt.Errorf(errDefault, err)
+       }
+
+       scheme.Default(objectTyped)
+       objectRaw, err := json.Marshal(objectTyped)
+       if err != nil {
+               return fmt.Errorf(errDefault, err)
+       }
+
+       objectMap := map[string]interface{}{}
+
+       err = json.Unmarshal(objectRaw, &objectMap)
+       if err != nil {
+               return fmt.Errorf(errDefault, err)
+       }
+
+       object.Object = objectMap
+
+       return nil
+}
```

mprahl:
@dhaiducek additionally, you'll need this in `main.go`:
```go
func init() {
	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
	utilruntime.Must(gktemplatesv1.AddToScheme(scheme))
	utilruntime.Must(gktemplatesv1beta1.AddToScheme(scheme))
	utilruntime.Must(v1.AddToScheme(eventsScheme))
	//+kubebuilder:scaffold:scheme
	utilruntime.Must(policiesv1.AddToScheme(scheme))
	utilruntime.Must(policiesv1.AddToScheme(eventsScheme))
	utilruntime.Must(extensionsv1.AddToScheme(scheme))
	utilruntime.Must(extensionsv1beta1.AddToScheme(scheme))
}
```

Those imports are:
```go
	gktemplatesv1 "github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1"
	gktemplatesv1beta1 "github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1beta1"
```

Additionally, you'll need to modify go.mod to include this requirement:
```
// This particular commit is used so that it doesn't force us to update k8s.io/api.
github.com/open-policy-agent/frameworks/constraint v0.0.0-20220504221328-9db6a6b27c97
```

controllers/templatesync/template_sync.go
```diff
@@ -335,28 +391,49 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
+		// Now that there is a mapping, there is a definitive answer whether it's cluster scoped rather
+		// than the previous educated guess based on the provided group and kind.
+		isClusterScoped = mapping.Scope == meta.RESTScopeRoot
+		resourceNs := ""
+
+		if !isClusterScoped {
+			resourceNs = instance.GetNamespace()
+		}
+
+		res = dClient.Resource(rsrc).Namespace(resourceNs)
```
JustinKuli, line 406: Woah, `.Namespace("")` works? I feel like we have something similar to what this is doing here in another controller... but we didn't do it this way. I don't know if either approach is better than the other...

JustinKuli: Config-policy-controller does this:
```
var res dynamic.ResourceInterface
if namespaced {
	res = r.TargetK8sDynamicClient.Resource(mapping.Resource).Namespace(object.Object.Metadata.Namespace)
} else {
	res = r.TargetK8sDynamicClient.Resource(mapping.Resource)
}
```

controllers/templatesync/template_sync.go
```diff
@@ -552,7 +675,28 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
+					tLogger.Error(err, fmt.Sprintf("Failed to apply %s label to CRD", utils.PolicyTypeLabel))
+				}
+
+				// The ConstraintTemplate does not generate status, so we need to generate an event for it
+				if isGkConstraintTemplate {
+					tLogger.Info("Emitting status event for " + gvk.Kind)
+					msg := fmt.Sprintf("%s %s was updated successfully", gvk.Kind, tName)
+					r.emitTemplateSuccess(instance, tIndex, tName, isClusterScoped, msg)
+				}
+			}
```
JustinKuli, line 697: Is this also at line 500?

JustinKuli: Ah, one is create, one is update. It's a bad code smell, but so is the fact that this Reconcile function is over 700 lines long...

Definitely something to address in the future.

controllers/templatesync/template_sync.go
JustinKuli: Having multi errors in go 1.20 will be nice

controllers/templatesync/template_sync.go
```diff
@@ -302,9 +350,17 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
-			errMsg := fmt.Sprintf("Mapping not found, please check if you have CRD deployed: %s", err)
+			errMsg := "Mapping not found, "
+			if isGkConstraintTemplate {
+				errMsg += "check if Gatekeeper is installed"
+			} else if gvk.Group == utils.GConstraint {
+				errMsg += "check if the required ConstraintTemplate has been deployed"
+			} else {
+				errMsg += "check if you have the CRD deployed"
+			}
+			errMsg += fmt.Sprintf(": %s", err)
```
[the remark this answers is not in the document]
dhaiducek: Haha! I noticed our docs don't say "please", either. Figured we might as well be consistent. 😄

controllers/templatesync/template_sync.go
dhaiducek: Yeah, I thought it wasn't working until I tried it again. I don't think I have a big preference--they're pretty similar.

JustinKuli:
/hold
/approve

It looks good to me, my comments do not require any changes. But it should probably get a final approval from @mprahl.

JustinKuli (approved):

controllers/templatesync/template_sync.go
dhaiducek: Definitely. There's a lot of replicated code--I tried to do my part to not make it too much worse by creating functions for repeated occurrences, but for sure could have gone a step further...

controllers/templatesync/template_sync.go
```diff
@@ -506,16 +621,24 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
-			r.emitTemplateError(instance, tIndex, tName, errMsg)
+			r.emitTemplateError(instance, tIndex, tName, isClusterScoped, errMsg)
 			tLogger.Error(resultError, "Failed to create the policy template")
 
 			policyUserErrorsCounter.WithLabelValues(instance.Name, tName, "format-error").Inc()
 
 			continue
 		}
 
+		// Fill in defaults set the by ConstraintTemplate CRD to ensure the spec comparison below is correct.
```
mprahl, line 634:
```suggestion
		// Fill in defaults set by the ConstraintTemplate CRD to ensure the spec comparison below is correct.
```

controllers/templatesync/template_sync.go
```diff
@@ -552,7 +675,28 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
-			err = r.handleSyncSuccess(ctx, instance, tIndex, tName, successMsg, res)
+			// Handle cluster scoped objects
+			if isClusterScoped {
+				addFinalizer = true
+
+				reqLogger.V(2).Info("Finalizer required for " + gvk.Kind)
+
+				// Ensure the CRD is labeled as policy-type=template so we can query custom resources for
+				// cleanup
+				err := applyCRDLabel(ctx, instance, tName, rsrc, dClient)
```
mprahl, line 686: Should we only set this when the policy template is Gatekeeper related even if it's cluster-scoped?

test/e2e/case17_gatekeeper_sync_test.go
```diff
@@ -0,0 +1,170 @@
...
+		By("Checking for the synced Constraint " + gkConstraintNameExtra)
+		expectedConstraint := propagatorutils.ParseYaml(gkConstraintYamlExtra)
+		Eventually(func() interface{} {
+			trustedPlc := propagatorutils.GetWithTimeout(clientManagedDynamic, gvrConstraint,
+				gkConstraintNameExtra, "", true, defaultTimeoutSeconds)
+
+			return trustedPlc.Object["spec"]
+		}, defaultTimeoutSeconds, 1).Should(propagatorutils.SemanticEqual(expectedConstraint.Object["spec"]))
+	})
+	It("should remove a Constraint when removed from policy-templates", func() {
```
mprahl, line 139: Could you also please cover the case when a ConstraintTemplate is removed?

test/e2e/case17_gatekeeper_sync_test.go
```diff
@@ -0,0 +1,170 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	propagatorutils "open-cluster-management.io/governance-policy-propagator/test/utils"
+
+	"open-cluster-management.io/governance-policy-framework-addon/test/utils"
+)
+
+var _ = Describe("Test Gatekeeper ConstraintTemplate sync", Ordered, Label("skip-minimum"), func() {
```
mprahl, line 20: I noticed that there is no coverage for the new logic in `overrideRemediationAction` but I'll cover this in my PR.

dhaiducek: Not complete coverage, but I do pass in `dryrun`, and it gets updated to `warn`...

mprahl: Good point!

controllers/templatesync/template_sync.go
dhaiducek: Right now, the controller watch list is filtered by the policy-template label, so we could restrict it to Gatekeeper-related objects, but I'm wondering now if we should have a policy allowlist and keep this particular check generic:
https://github.com/open-cluster-management-io/governance-policy-framework-addon/blob/1048a8040d8a20bddec82b1afed3f89a26560198/main.go#L339-L351

You're bringing up a good point that now that I've added CRD permissions, there's zero filtering for the template-sync passing along a CRD. (@JustinKuli tested it syncing an Event since it has those permissions--and it synced it.)

controllers/templatesync/template_sync.go
dhaiducek: Oof. You got me. 😉

dhaiducek:
With the list on ConstraintTemplates, I'm now seeing "failed to list" messages:
```
2023-03-07T16:02:06.873-0500	info	klog	cache/reflector.go:324	pkg/mod/k8s.io/client-go@v0.23.10/tools/cache/reflector.go:167: failed to list *v1.ConstraintTemplate: the server could not find the requested resource (get constrainttemplates.templates.gatekeeper.sh)
2023-03-07T16:02:06.873-0500	error	klog	cache/reflector.go:138	pkg/mod/k8s.io/client-go@v0.23.10/tools/cache/reflector.go:167: Failed to watch *v1.ConstraintTemplate: failed to list *v1.ConstraintTemplate: the server could not find the requested resource (get constrainttemplates.templates.gatekeeper.sh)

k8s.io/client-go/tools/cache.DefaultWatchErrorHandler
	/Users/dhaiduce/go/pkg/mod/k8s.io/client-go@v0.23.10/tools/cache/reflector.go:138
k8s.io/client-go/tools/cache.(*Reflector).Run.func1
	/Users/dhaiduce/go/pkg/mod/k8s.io/client-go@v0.23.10/tools/cache/reflector.go:222
k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1
	/Users/dhaiduce/go/pkg/mod/k8s.io/apimachinery@v0.23.10/pkg/util/wait/wait.go:155
k8s.io/apimachinery/pkg/util/wait.BackoffUntil
	/Users/dhaiduce/go/pkg/mod/k8s.io/apimachinery@v0.23.10/pkg/util/wait/wait.go:156
k8s.io/client-go/tools/cache.(*Reflector).Run
	/Users/dhaiduce/go/pkg/mod/k8s.io/client-go@v0.23.10/tools/cache/reflector.go:220
k8s.io/apimachinery/pkg/util/wait.(*Group).StartWithChannel.func1
	/Users/dhaiduce/go/pkg/mod/k8s.io/apimachinery@v0.23.10/pkg/util/wait/wait.go:56
k8s.io/apimachinery/pkg/util/wait.(*Group).Start.func1
	/Users/dhaiduce/go/pkg/mod/k8s.io/apimachinery@v0.23.10/pkg/util/wait/wait.go:73
```
They seem to appear after the ConstraintTemplate CRD is added and then removed but not before then, so maybe it's not a big deal? I'm not sure how to remove the informer or prevent these from appearing, though.

dhaiducek:
The tests passed consistently after some E2E tweaking. Though I'm a little confused on why the `minimum` tests would take twice as long on the two most recent runs (28 min vs 57 min), but maybe it's something environmental?

mprahl:
> The tests passed consistently after some E2E tweaking. Though I'm a little confused on why the `minimum` tests would take twice as long on the two most recent runs (28 min vs 57 min), but maybe it's something environmental?

@dhaiducek was it reasonably fast when running locally with minimum?

controllers/templatesync/template_sync.go
```diff
@@ -316,15 +379,47 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
+			policySystemErrorsCounter.WithLabelValues(request.Name, tName, "get-error").Inc()
+
+			// The CRD should exist since it was found in the mapping previously; Requeue this template
+			return reconcile.Result{}, err
+		}
+
+		// If no policy-type=template label AND the GroupKind is not on the explicit allow list, don't
+		// sync this template
+		if !hasTemplateLabel && !utils.IsAllowedPolicy(gvk.GroupKind()) {
+			errMsg := fmt.Sprintf("API is not supported: %s", gvk.String())
```
mprahl, line 397:
```suggestion
			errMsg := fmt.Sprintf("the policy-template kind is not supported: %s", gvk.String())
```

controllers/templatesync/template_sync.go
```diff
@@ -316,15 +379,47 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
+			err := fmt.Errorf(errMsg)
+
+			resultError = err
+
+			r.emitTemplateError(instance, tIndex, tName, isClusterScoped, errMsg)
+			tLogger.Error(err, "Unsupported API found in object definition",
```
mprahl, line 403:
```suggestion
			tLogger.Error(err, "Unsupported policy-template found in object definition",
```

controllers/templatesync/template_sync.go
```diff
@@ -866,3 +1206,116 @@ func getLatestStatusMessage(pol *policiesv1.Policy, tIndex int) string {
...
+		if k8serrors.IsNotFound(err) {
+			return false, nil
+		}
+
+		return false, err
+	}
+
+	crdLabels := crd.GetLabels()
+	if crdLabels == nil {
+		crdLabels = map[string]string{}
```
mprahl, line 1232: You should be able to read a nil map just fine so I would skip this.

controllers/templatesync/template_sync.go
```diff
@@ -866,3 +1206,116 @@ func getLatestStatusMessage(pol *policiesv1.Policy, tIndex int) string {
...
+	}
+
+	// Check for the policy-type=template label
+	if val, ok := crdLabels[utils.PolicyTypeLabel]; ok {
+		if val == "template" {
+			return true, nil
+		}
+	}
+
+	return false, nil
```
mprahl, line 1242:
```suggestion
	return crd.GetLabels()[utils.PolicyTypeLabel] == "template", nil
```

mprahl:
@dhaiducek nice work! I just left a few minor comments. As long as we confirm that the tests run quickly locally on "minimum", this should be good to merge after those minor comments are addressed.

controllers/templatesync/template_sync.go
dhaiducek: Ah, I didn't know an empty map could tolerate being read.

So I guess this `if` can be removed also, then:
```go
// Handle adding metadata labels
labels := tObjectUnstructured.GetLabels()
if labels == nil {
  labels = map[string]string{}
}
```
https://github.com/dhaiducek/governance-policy-framework-addon/blob/7062a5ea54db942ad555dc7d8562f07d8ae5a8ea/controllers/templatesync/template_sync.go#L501-L503

mprahl: @dhaiducek you'll need to keep that one because you can't write to a nil map.

mprahl (dismissed):
Nice work!

dhaiducek:
/unhold

Confirmed that it takes about 20 minutes locally, but now it looks like there might still be some flakes. I'll poke at it for a couple minutes this afternoon before re-running them.

JustinKuli (approved):

PR #34, opened 2023-02-23
Remove the dependency watches when a policy is deleted

mprahl:
This will prevent unnecessary reconciles for dependency updates after the policy that defined the dependencies is deleted.


controllers/templatesync/template_sync.go
```diff
@@ -87,6 +87,13 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 
 	// Fetch the Policy instance
 	instance := &policiesv1.Policy{}
+	policyObjectID := depclient.ObjectIdentifier{
```
JustinKuli, line 90: Moving this 500 lines is just a reminder that this function is getting too long

JustinKuli (approved):

PR #35, opened 2023-03-01
Set parent policy label on template when updating

willkutler:
Currently, the parent policy label on templates, which is used for https://github.com/open-cluster-management-io/governance-policy-framework-addon/pull/28 , is only set on template creation, which could cause issues with policies that have already been created. This fix verifies the necessary template labels are set when updating the template. It also includes a light refactor of the template equality check to make it look a little cleaner.

### DIFF
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -666,19 +666,23 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 				log.Error(err, "Failed to apply defaults to the ConstraintTemplate for comparison. Continuing.")
 			}
 		}
+		// verify parent policy label is set properly
+		tObjectUnstructured.SetLabels(r.setDefaultTemplateLabels(instance, tObjectUnstructured.GetLabels()))
 
 		overrideRemediationAction(instance, tObjectUnstructured)
+
 		// got object, need to compare both spec and annotation and update
 		eObjectUnstructured := eObject.UnstructuredContent()
-		if (!equality.Semantic.DeepEqual(eObjectUnstructured["spec"], tObjectUnstructured.Object["spec"])) ||
-			(!equality.Semantic.DeepEqual(eObject.GetAnnotations(), tObjectUnstructured.GetAnnotations())) ||
-			(!equality.Semantic.DeepEqual(eObject.GetOwnerReferences(), tObjectUnstructured.GetOwnerReferences())) {
+
+		if !equivalentTemplates(eObject, tObjectUnstructured) {
 			// doesn't match
 			tLogger.Info("Existing object and template didn't match, will update")
 
 			eObjectUnstructured["spec"] = tObjectUnstructured.Object["spec"]
 
 			eObject.SetAnnotations(tObjectUnstructured.GetAnnotations())
+
+			eObject.SetLabels(tObjectUnstructured.GetLabels())
 
 			eObject.SetOwnerReferences(tObjectUnstructured.GetOwnerReferences())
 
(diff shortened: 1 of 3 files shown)


controllers/templatesync/template_sync.go
```diff
@@ -605,6 +610,51 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
+		parentPolicyLabel:            instance.GetName(),
+		"cluster-name":               instance.GetLabels()[common.ClusterNameLabel],
+		common.ClusterNameLabel:      instance.GetLabels()[common.ClusterNameLabel],
+		"cluster-namespace":          instance.GetLabels()[common.ClusterNamespaceLabel],
+		common.ClusterNamespaceLabel: instance.GetLabels()[common.ClusterNamespaceLabel],
+	}
+
+	for key, label := range desiredLabels {
+		_, ok := labels[key]
+		if !ok {
```
mprahl, line 650: It might be better to always set these desired labels in the event a user manually changed one of the labels in the `policy-templates` array.

controllers/templatesync/template_sync.go
```diff
@@ -605,6 +610,51 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
+
+	if !equality.Semantic.DeepEqual(eObject.GetOwnerReferences(), tObject.GetOwnerReferences()) {
+		return false
+	}
+
+	return true
+}
+
+// verifyTemplateLabels ensures the template contains all necessary labels for processing
+func verifyTemplateLabels(instance *policiesv1.Policy, labels map[string]string) map[string]string {
```
mprahl, line 635: Would `setDefaultTemplateLabels` be a better name?

controllers/templatesync/template_sync.go
```diff
@@ -605,6 +610,48 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
+func setDefaultTemplateLabels(instance *policiesv1.Policy, labels map[string]string) map[string]string {
+	if labels == nil {
+		labels = map[string]string{}
+	}
+
+	desiredLabels := map[string]string{
+		parentPolicyLabel:            instance.GetName(),
+		"cluster-name":               instance.GetLabels()[common.ClusterNameLabel],
+		common.ClusterNameLabel:      instance.GetLabels()[common.ClusterNameLabel],
+		"cluster-namespace":          instance.GetLabels()[common.ClusterNamespaceLabel],
```
dhaiducek, line 644: @mprahl Will this battle with the propagator in hosted mode?

willkutler: I updated it - is it just the cluster-namespace label or the one below it as well?

dhaiducek: I think it'd be both, but I'll let @mprahl answer since I don't really have a good handle on hosted mode yet...

But I'm also wondering whether there should be a specialized check for this label in the `EquivalentTemplates` function for hosted mode so that it's not updated every time?

willkutler: @mprahl could I get an update on this?

controllers/templatesync/template_sync.go
```diff
@@ -605,6 +610,50 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
+	if labels == nil {
+		labels = map[string]string{}
+	}
+
+	desiredLabels := map[string]string{
+		parentPolicyLabel:            instance.GetName(),
+		"cluster-name":               instance.GetLabels()[common.ClusterNameLabel],
+		common.ClusterNameLabel:      instance.GetLabels()[common.ClusterNameLabel],
+		"cluster-namespace":          r.ClusterNamespace,
+		common.ClusterNamespaceLabel: instance.GetLabels()[common.ClusterNamespaceLabel],
```
mprahl, line 647:
```suggestion
		common.ClusterNamespaceLabel: r.ClusterNamespace,
```

controllers/templatesync/template_sync.go
```diff
@@ -504,8 +504,8 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 
 				labels["cluster-name"] = instance.GetLabels()[common.ClusterNameLabel]
```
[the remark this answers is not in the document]
willkutler: done!

mprahl (approved):

PR #36, opened 2023-03-13
Add the Gatekeeper constraint status sync controller

mprahl:
This controller is responsible for relaying Gatekeeper constraint audit results as policy status events for constraints associated with a policy.

The controller will only start if Gatekeeper is installed. If Gatekeeper's installation status changes while the controller manager is running, the health endpoint will return as unhealthy to cause Kubernetes to restart the pod to then enable/disable this new controller.

Relates:
https://issues.redhat.com/browse/ACM-3328

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -283,6 +283,10 @@ install-resources:
 	-kubectl create ns $(MANAGED_CLUSTER_NAME) --kubeconfig=$(HUB_CONFIG)
 	@echo creating namespace on managed
 	-kubectl create ns $(MANAGED_CLUSTER_NAME) --kubeconfig=$(MANAGED_CONFIG)
+	@if [ "$(KIND_VERSION)" != "minimum" ]; then \
+		echo installing Gatekeeper on the managed cluster; \
+		kubectl apply -f https://raw.githubusercontent.com/open-policy-agent/gatekeeper/v3.11.0/deploy/gatekeeper.yaml  --kubeconfig=$(MANAGED_CONFIG); \
+	fi
 
 .PHONY: e2e-dependencies
 e2e-dependencies:
--- a/controllers/gatekeepersync/gatekeeper_constraint_sync.go
+++ b/controllers/gatekeepersync/gatekeeper_constraint_sync.go
@@ -1,0 +1,426 @@
+// Copyright Contributors to the Open Cluster Management project
+
+package gatekeepersync
+
+import (
+	"context"
+	"crypto/sha1"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+	"sync"
+
+	depclient "github.com/stolostron/kubernetes-dependency-watches/client"
+	admissionregistration "k8s.io/api/admissionregistration/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/dynamic"
+	policyv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	"open-cluster-management.io/governance-policy-framework-addon/controllers/utils"
+)
+
+const (
+	ControllerName        = "gatekeeper-constraint-status-sync"
+	GatekeeperWebhookName = "gatekeeper-validating-webhook-configuration"
+)
+
+var log = logf.Log.WithName(ControllerName)
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *GatekeeperConstraintReconciler) SetupWithManager(mgr ctrl.Manager, constraintEvents *source.Channel) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&policyv1.Policy{}).
+		WithEventFilter(policyPredicates()).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 5}).
+		Watches(constraintEvents, &handler.EnqueueRequestForObject{}).
+		Named(ControllerName).
+		Complete(r)
+}
+
+// blank assignment to verify that ReconcilePolicy implements reconcile.Reconciler
+var _ reconcile.Reconciler = &GatekeeperConstraintReconciler{}
+
+// Used to track sent messages for a particular Gatekeeper constraint.
+type policyKindName struct {
+	Policy string
+	Kind   string
+	Name   string
+}
+
+// GatekeeperConstraintReconciler is responsible for relaying Gatekeeper constraint audit results as policy status
+// events.
+type GatekeeperConstraintReconciler struct {
+	client.Client
+	utils.ComplianceEventSender
+	Scheme             *runtime.Scheme
+	DynamicClient      dynamic.Interface
+	ConstraintsWatcher depclient.DynamicWatcher
+	// A cache of sent messages to avoid repeating status events due to race conditions. Each value is a SHA1
+	// digest.
+	lastSentMessages sync.Map
+}
+
+//+kubebuilder:rbac:groups=policy.open-cluster-management.io,resources=policies,verbs=get;list;watch
+//+kubebuilder:rbac:groups=constraints.gatekeeper.sh,resources=*,verbs=create;get;list;watch
+//+kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=validatingwebhookconfigurations,resourceNames=gatekeeper-validating-webhook-configuration,verbs=get;list;watch
+//+kubebuilder:rbac:groups=core,resources=events,verbs=create;delete;get;list;patch;update;watch
+
+// Reconcile handles Policy objects that contain a Gatekeeper constraint and relays status messages from Gatekeeper
+// audit results. Every time a Gatekeeper constraint in a Policy is updated, a reconcile on the Policy is triggered.
+func (r *GatekeeperConstraintReconciler) Reconcile(
+	ctx context.Context, request reconcile.Request,
+) (
+	reconcile.Result, error,
+) {
+	log := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	log.Info("Reconciling a Policy with one or more Gatekeeper constraints")
+
+	policyObjID := depclient.ObjectIdentifier{
+		Group:     policyv1.GroupVersion.Group,
+		Version:   policyv1.GroupVersion.Version,
+		Kind:      "Policy",
+		Namespace: request.Namespace,
+		Name:      request.Name,
+	}
+	policy := &policyv1.Policy{}
+
+	err := r.Get(ctx, request.NamespacedName, policy)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			log.V(1).Info("The Policy was deleted. Cleaning up watchers and status message cache.")
+
+			r.lastSentMessages.Range(func(key, value any) bool {
+				keyTyped := key.(policyKindName)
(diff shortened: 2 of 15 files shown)


main.go
```diff
@@ -363,6 +371,89 @@ func getManager(
...
+		"governance-policy-framework-addon", tool.Options.HubConfigFilePathName,
+	)
+	if err != nil {
+		log.Error(err, "unable to setup a configChecker")
+		os.Exit(1)
+	}
+
+	healthCheck := configChecker.Check
+
+	// Gatekeeper is not supported on such old versions of Kubernetes, so just always disable it in this case.
```
JustinKuli, line 393: I think this comment is confusing, since we're not conditionally *disabling* it, we're conditionally *enabling* it.
```suggestion
	// Only enable the GatekeeperConstraintReconciler if the Kubernetes version is v1.16 or above
```

controllers/gatekeepersync/health.go
```diff
@@ -0,0 +1,95 @@
...
+) {
+	fieldSelector := fmt.Sprintf("metadata.name=constrainttemplates.%s", utils.GvkConstraintTemplate.Group)
+	timeout := int64(30)
+	crdGVR := schema.GroupVersionResource{
+		Group:    "apiextensions.k8s.io",
+		Version:  "v1",
+		Resource: "customresourcedefinitions",
+	}
+
+	listResult, err := dynamicClient.Resource(crdGVR).List(
```
JustinKuli, line 39: If this is the first thing done by the dynamic client (and I think it might be) then I think it could take a long time to do its API discovery... you might already know more about that than me, so correct me if we don't need to be concerned with that here. But if this `List` times out (because of that issue), then the error would bubble up to `main`, and we'd effectively restart the container. But restarting the container will just cause it to have the same problem again, unless I'm missing something.

controllers/gatekeepersync/health.go
```diff
@@ -0,0 +1,95 @@
...
+		if lastHealthErr != nil {
+			return lastHealthErr
+		}
+
+		select {
+		case <-w.Done():
+			select {
+			case <-ctx.Done():
+				// Stop the retry watcher if the parent context is canceled.
+				w.Stop()
```
JustinKuli, line 72: If the parent context was canceled, should we return a non-nil error? Otherwise it feels like there *could* be some weird situation where it never appears unhealthy, because it keeps hitting this spot?

controllers/gatekeepersync/gatekeeper_constraint_sync.go
```diff
@@ -0,0 +1,434 @@
...
+			r.lastSentMessages.Range(func(key, value any) bool {
+				keyTyped := key.(policyKindName)
+				if keyTyped.Policy == request.Name {
+					r.lastSentMessages.Delete(keyTyped)
+				}
+
+				return true
+			})
+
+			err := r.ConstraintsWatcher.RemoveWatcher(policyObjID)
```
JustinKuli, line 132: I believe this only returns a non-nil error if the policyObjID is invalid... it shouldn't be invalid, so it should always be nil, but if it *was* somehow invalid then the error here would cause the policy to be re-queued forever.

JustinKuli (changes requested):
This is only a partial review, but I had some early questions.

dhaiducek (commented):
I gave this a first pass, and I don't think I saw anything glaring. I'll look at it again first thing tomorrow when my vviissiioonn isn't blurring. 😉

Makefile
```diff
@@ -283,6 +283,10 @@ install-resources:
 	-kubectl create ns $(MANAGED_CLUSTER_NAME) --kubeconfig=$(HUB_CONFIG)
 	@echo creating namespace on managed
 	-kubectl create ns $(MANAGED_CLUSTER_NAME) --kubeconfig=$(MANAGED_CONFIG)
+	@if [ "$(KIND_VERSION)" != "minimum" ]; then \
+		echo installing Gatekeeper on the managed cluster; \
+		kubectl apply -f https://raw.githubusercontent.com/open-policy-agent/gatekeeper/v3.11.0/deploy/gatekeeper.yaml  --kubeconfig=$(MANAGED_CONFIG); \
+	fi
```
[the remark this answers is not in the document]
dhaiducek: It follows from technical debt in my PR:
> Handling for older Kubernetes/older Gatekeeper (right now, I have a skip-minimum tag that skips the test entirely); Gatekeeper v3.4 added a PodDisruptionBudget that isn't available until Kubernetes v1.21

https://github.com/open-cluster-management-io/governance-policy-framework-addon/pull/33#issuecomment-1450652169

test/e2e/case17_gatekeeper_sync_test.go
```diff
@@ -42,31 +51,102 @@ var _ = Describe("Test Gatekeeper ConstraintTemplate sync", Ordered, Label("skip
...
+			var available bool
+			for _, deployment := range deployments.Items {
+				for _, condition := range deployment.Status.Conditions {
+					if condition.Reason == "MinimumReplicasAvailable" {
+						available = condition.Status == "True"
+					}
+				}
+				g.Expect(available).To(BeTrue())
+			}
+		}, defaultTimeoutSeconds, 1).Should(Succeed())
```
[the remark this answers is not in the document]
mprahl: Good catch.

controllers/utils/events.go
```diff
@@ -0,0 +1,95 @@
...
+	policyv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type EventSender interface {
+	GetInstanceName() string
+	GetClusterNamespace() string
+	GetClientSet() *kubernetes.Clientset
+	GetControllerName() string
+}
```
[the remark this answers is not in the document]
mprahl: I felt like it was clunky as well. I'm going to embed a struct with a `SendEvent` method that I think will be cleaner.

controllers/utils/events.go
```diff
@@ -0,0 +1,95 @@
...
+
+// SendComplianceEvent will send a policy template status message update synchronously as opposed to EventRecorder
+// sending events in the background asynchronously.
+func SendComplianceEvent(
+	ctx context.Context,
+	e EventSender,
+	instance client.Object,
+	owner metav1.OwnerReference,
```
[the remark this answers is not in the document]
mprahl: I think it's good enough for the moment. If you come up with an alternative, please let me know.

main.go
mprahl: I think I'll remove this semver check and just assume if Gatekeeper is installed, it is working.

controllers/gatekeepersync/health.go
mprahl: The dynamic client doesn't perform a discovery from what I can tell in the code. It just uses the GVR to construct the Kubernetes API endpoint URL.

controllers/gatekeepersync/health.go
mprahl: Yes, good point.

controllers/gatekeepersync/gatekeeper_constraint_sync.go
```diff
@@ -0,0 +1,434 @@
...
+		APIVersion: refreshedPolicy.APIVersion,
+		Kind:       refreshedPolicy.Kind,
+		Name:       refreshedPolicy.Name,
+		UID:        refreshedPolicy.UID,
+	}
+
+	if len(refreshedPolicy.Status.Details) < templateIndex+1 ||
+		len(refreshedPolicy.Status.Details[templateIndex].History) == 0 ||
+		refreshedPolicy.Status.Details[templateIndex].History[0].Message != fmt.Sprintf("%s; %s", compliance, msg) {
+		err = utils.SendComplianceEvent(ctx, r, constraint, owner, msg, compliance)
```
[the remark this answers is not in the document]
mprahl: I'll rework this so that if it attempts to send the same status message and it's already in the Policy status, the entry in the lastSentMessages is deleted.

test/e2e/case17_gatekeeper_sync_test.go
```diff
@@ -161,6 +245,205 @@ var _ = Describe("Test Gatekeeper ConstraintTemplate sync", Ordered, Label("skip
...
+
+			history := managedPolicy.Status.Details[1].History
+			g.Expect(len(history)).ToNot(Equal(0))
+			expectedMsg := `NonCompliant; The Gatekeeper validating webhook is disabled but the constraint's ` +
+				`spec.enforcementAction is deny. deny - All configmaps must have a 'my-gk-test' label (on ConfigMap ` +
+				`case17-gk-test/case17-test); deny - All configmaps must have a 'my-gk-test' label (on ConfigMap ` +
+				`case17-gk-test/case17-test2)`
+			g.Expect(history[0].Message).To(
+				Equal(expectedMsg),
+				fmt.Sprintf("Got %s but expected %s", history[0].Message, expectedMsg),
```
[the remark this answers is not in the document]
mprahl: It has a similar message but it truncates it so much that it's hard to debug.

controllers/gatekeepersync/policy_predicate.go
```diff
@@ -0,0 +1,38 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+// policyPredicates filters out policies without Gatekeeper constraints and policy updates without the generation
+// changing.
+func policyPredicates() predicate.Funcs {
+	return predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			// nolint: forcetypeassert
```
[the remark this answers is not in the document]
mprahl: It looks like it wasn't needed. I guess it was preemptive on my part.  :laughing:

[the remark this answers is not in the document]
JustinKuli: Maybe we need to re-enable the linter somewhere? Because I would rather not have panics 😄

JustinKuli (dismissed):
/hold

Looks good to me! Thanks for those updates. Leaving it up to @dhaiducek to double-check :)

controllers/gatekeepersync/gatekeeper_constraint_sync.go
```diff
@@ -0,0 +1,426 @@
...
+)
+
+var log = logf.Log.WithName(ControllerName)
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *GatekeeperConstraintReconciler) SetupWithManager(mgr ctrl.Manager, constraintEvents *source.Channel) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&policyv1.Policy{}).
+		WithEventFilter(policyPredicates()).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 5}).
```
dhaiducek, line 47: There isn't concurrency in the other controllers that I can see. Concurrency seems good, but I'm wondering about the effect here.

test/e2e/case17_gatekeeper_sync_test.go
```diff
@@ -125,6 +208,205 @@ var _ = Describe("Test Gatekeeper ConstraintTemplate sync", Ordered, Label("skip
...
+			g.Expect(managedPolicy.Status.Details).To(HaveLen(2))
+			g.Expect(managedPolicy.Status.Details[1].TemplateMeta.GetName()).To(Equal(gkConstraintName))
+			history := managedPolicy.Status.Details[1].History
+			g.Expect(len(history)).ToNot(Equal(0))
+			expectedMsg := "Compliant; The constraint has no violations"
+			g.Expect(history[0].Message).To(
+				Equal(expectedMsg),
+				fmt.Sprintf("Got %s but expected %s", history[0].Message, expectedMsg),
+			)
+		}, "120s", 1).Should(Succeed())
```
dhaiducek, line 234: Nit: Following other conventions here. Is there a reason to hardcode this value (as well as the other instances in this suite)?
```suggestion
		}, defaultTimeoutSeconds*4, 1).Should(Succeed())
```

dhaiducek (approved):
LGTM!
I have a musing about concurrency here where it doesn't seem to be elsewhere and using `defaultTimeoutSeconds` consistently, but they're not anything that should block the PR.

controllers/gatekeepersync/gatekeeper_constraint_sync.go
mprahl: @dhaiducek you're right. I added concurrency here because by default, the Gatekeeper audits run every minute so if the user has a lot of constraints, I didn't want there to be a delay in getting results.

The way the concurrency works in controller-runtime is that there is only one goroutine processing a particular reconcile request at a time. For example, since we are reconciling `Policy` objects, if there are three reconciles requests for `Policy` named `a` and one reconcile request for `Policy` named `b`, there would just be one goroutine processing `Policy` `a` and one goroutine processing `Policy` `b`.

In other words, it's concurrent across reconciles for different `Policy` objects but reconciles for the same `Policy` will not run concurrently.

test/e2e/case17_gatekeeper_sync_test.go
mprahl: I changed this to use a different constant.

PR #38, opened 2023-03-21
Ignore gosec failures for SHA1 usage

mprahl:
The SHA1 hashes are used for uniqueness checks in a cache. They are not used for anything security related.

### DIFF
--- a/controllers/gatekeepersync/gatekeeper_constraint_sync.go
+++ b/controllers/gatekeepersync/gatekeeper_constraint_sync.go
@@ -4,6 +4,8 @@ package gatekeepersync
 
 import (
 	"context"
+
+	// #nosec G505
 	"crypto/sha1"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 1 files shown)


controllers/gatekeepersync/gatekeeper_constraint_sync.go
```diff
@@ -345,6 +347,7 @@ func (r *GatekeeperConstraintReconciler) sendComplianceEvent(
 	if len(refreshedPolicy.Status.Details) < templateIndex+1 ||
 		len(refreshedPolicy.Status.Details[templateIndex].History) == 0 ||
 		refreshedPolicy.Status.Details[templateIndex].History[0].Message != fmt.Sprintf("%s; %s", compliance, msg) {
+		//#nosec G401
```
dhaiducek, line 350: Nit: Because it looks prettier
```suggestion
		// #nosec G401
```

dhaiducek (approved):
/hold if you agree with my change. Otherwise, merge away! ⛵

controllers/gatekeepersync/gatekeeper_constraint_sync.go
mprahl: @JustinKuli mentioned that human comments have spaces after `//` but ones for tools don't. Am I characterizing that right @JustinKuli ?

mprahl:
I'll rerun the test, but I'll investigate the flakiness for a separate PR.

controllers/gatekeepersync/gatekeeper_constraint_sync.go
dhaiducek: Ah, you're right: https://github.com/securego/gosec#annotating-code
That's lame.

mprahl:
> I'll rerun the test, but I'll investigate the flakiness for a separate PR.

I can't reproduce the flaky test locally. I reran the Gatekeeper test several times simulating hosted mode.

controllers/gatekeepersync/gatekeeper_constraint_sync.go
JustinKuli: I don't remember when I mentioned it, but yeah, I think the general idea is that comments that will be generated or read by machines will not have a space (eg, `go:generate`, `+kubebuilder`, and pragmas). But it's not a strict rule in most cases, and I don't think we're perfectly consistent with the kubebuilder comments anyway

PR #39, opened 2023-03-22
Perform discovery API calls only as needed

mprahl:
This commit changes the discovery REST mapper that was used to be targeted to
that specific resource version.

This is to reduce load on the Kubernetes API as well as to reduce the chance
of long delays due to queries per second (QPS) exceeding the default limit.

An additional commit was added to prevent watches being created on Gatekeeper constraints without the corresponding CRD installed.

Other commits are to address flaky tests.

Relates:
https://issues.redhat.com/browse/ACM-4518

### DIFF
--- a/controllers/gatekeepersync/gatekeeper_constraint_sync.go
+++ b/controllers/gatekeepersync/gatekeeper_constraint_sync.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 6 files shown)


mprahl:
@dhaiducek and @JustinKuli I think this should help with the flaky tests.

mprahl:
/unhold

@JustinKuli and @dhaiducek I figured out a better approach.

mprahl:
@dhaiducek and @JustinKuli sorry for the noise. This is now ready for review.

JustinKuli (commented):
How will we measure if this is better than the old implementation?

JustinKuli (approved):

PR #41, opened 2023-03-27
Disable the Gatekeeper controller when run in old Kubernetes clusters

mprahl:
Relates:
https://issues.redhat.com/browse/ACM-4588

### DIFF
--- a/main.go
+++ b/main.go
@@ -371,14 +371,6 @@ func getManager(
 		os.Exit(1)
 	}
 
-	discoveryClient := discovery.NewDiscoveryClientForConfigOrDie(managedCfg)
-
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -369,6 +371,14 @@ func getManager(
 		os.Exit(1)
 	}
 
+	discoveryClient := discovery.NewDiscoveryClientForConfigOrDie(managedCfg)
+
+	serverVersion, err := discoveryClient.ServerVersion()
+	if err != nil {
```
JustinKuli, line 377: I took a quick look and I think this is true... this won't do an api-resources refresh or whatever, that could take a long time, right?

JustinKuli (approved):

PR #43, opened 2023-03-31
Conditionally clean up Gatekeeper Constraints

dhaiducek:
Listing Gatekeeper Constraints was refreshing the discovery API. This adds a global boolean to only attempt cleanup when a Constraint was previously synced successfully.

### DIFF
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -73,11 +73,12 @@ type PolicyReconciler struct {
 	// This client, initialized using mgr.Client() above, is a split client
 	// that reads objects from the cache and writes to the apiserver
 	client.Client
-	DynamicWatcher   depclient.DynamicWatcher
-	Scheme           *runtime.Scheme
-	Config           *rest.Config
-	Recorder         record.EventRecorder
-	ClusterNamespace string
+	DynamicWatcher      depclient.DynamicWatcher
+	Scheme              *runtime.Scheme
+	Config              *rest.Config
+	Recorder            record.EventRecorder
+	ClusterNamespace    string
+	createdGkConstraint *bool
 }
 
 // Reconcile reads that state of the cluster for a Policy object and makes changes based on the state read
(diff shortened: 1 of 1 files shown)


JustinKuli (dismissed):
/hold for @mprahl 

LGTM! It's easier than I would have expected to transition the global back to false, so that's really nice.

controllers/templatesync/template_sync.go
```diff
@@ -47,7 +47,10 @@ const (
 	ControllerName string = "policy-template-sync"
 )
 
-var log = ctrl.Log.WithName(ControllerName)
+var (
+	createdGkConstraint = false
```
mprahl, line 51: Could you make this a struct member of the reconciler instead?

controllers/templatesync/template_sync.go
```diff
@@ -690,6 +695,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 			// Handle cluster scoped objects
 			if isClusterScoped {
 				addFinalizer = true
+				createdGkConstraint = true
```
mprahl, line 698: It'd be nice to not assume that any cluster scoped policy template is Gatekeeper.

controllers/templatesync/template_sync.go
```diff
@@ -854,51 +866,29 @@ func (r *PolicyReconciler) cleanUpExcessTemplates(
...
 						Group:    utils.GvkConstraintTemplate.Group,
 						Resource: "constrainttemplates",
-						Version:  "v1beta1",
+						Version:  "v1",
 					},
 					namespaced: false,
 				})
+			} else {
+				// No ConstraintTemplates found -- reset boolean
+				*r.createdGkConstraint = false
```
mprahl, line 888: You'll want to explicitly set this to `false` on errors when the CRD doesn't exist error is encountered.

mprahl (approved):

mprahl (changes requested):
Just one more comment.

controllers/templatesync/template_sync.go
```diff
@@ -908,12 +898,45 @@ func (r *PolicyReconciler) cleanUpExcessTemplates(
...
+				}
+				// Log and ignore other errors to allow cleanup to continue since Gatekeeper may not be installed
+			} else {
+				reqLogger.Info(fmt.Sprintf("Ignoring ConstraintTemplate cleanup error: %s", err.Error()))
+				r.setCreatedGkConstraint(false)
+			}
 		} else {
 			reqLogger.Info(fmt.Sprintf("Ignoring ConstraintTemplate cleanup error: %s", err.Error()))
+			r.setCreatedGkConstraint(false)
 		}
```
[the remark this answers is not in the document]
dhaiducek: Oh yeah, that makes more sense.

PR #44, opened 2023-04-03
3329 finalizer cleanup

JustinKuli:
A new "controller" watches the Deployment for an uninstall annotation. When the annotation is true, it updates a flag so that all the other controllers don't have to track it individually. Most controllers just skip their reconciles during the uninstall, but the template-sync will clean up its finalizers (and skip an error case during that process).

The default behavior is still to run the controllers. But if the first command line argument is `trigger-uninstall`, then the process will add the uninstallation annotation to the controller's deployment, and delete all the policies on the cluster. The main intent of this is to be run as a pre-delete hook when uninstalling the addon.

https://issues.redhat.com/browse/ACM-3329

### DIFF
--- a/.github/workflows/kind.yml
+++ b/.github/workflows/kind.yml
@@ -79,16 +79,21 @@ jobs:
         export COVERAGE_E2E_OUT=coverage_e2e_hosted_mode.out
         make e2e-test-coverage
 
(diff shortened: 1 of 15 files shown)


mprahl (approved):
Nice!

PR #46, opened 2023-04-04
Make conflicted object template error more readable

willkutler:
Replaces the error returned by the CRD for conflicting object-templates and object-templates-raw with a more readable one. There does not seem to be a way to add a custom error message for this failure in the CRD itself, so this change just watches for the error message and replaces it with a more readable one before the event is sent.

ref: https://issues.redhat.com/browse/ACM-4083


JustinKuli (approved):
Nice!

PR #47, opened 2023-04-04
Stop resetting the policy template compliance when already reset

mprahl:
When it's already reset or is empty, the patch fails and leads to the reconcile request being retried until the corresponding controller that provides status for the policy template.

Relates:
https://issues.redhat.com/browse/ACM-4486


JustinKuli (approved):
LGTM

PR #48, opened 2023-04-04
Stop updating on every reconcile when pruneObjectBehavior is not set

mprahl:
equivalentTemplates would always return false when a ConfigurationPolicy was encountered without pruneObjectBehavior explicitly set since the Kubernetes API returns "None" as the default value.

This isn't an ideal solution, but it's inexpensive compared to performing API queries to see what defaults are and this is a common case.


gparvin (approved):
lgtm

PR #51, opened 2023-04-07
Get deployment name from environment variable

JustinKuli:
The environment variable will be populated from the labels on the pod, using the downward API.

Refs:
 - https://issues.redhat.com/browse/ACM-3329

Fixes problems found in https://github.com/open-cluster-management-io/governance-policy-addon-controller/pull/71, where the deployment name is different [in the addon-controller](https://github.com/open-cluster-management-io/governance-policy-addon-controller/blob/cce56fadef78a48b33524d1c99888de59c7ec7d7/pkg/addon/policyframework/manifests/managedclusterchart/templates/deployment.yaml#L6) than here... Looking back, hard-coding it here was not the best idea I've ever had.


dhaiducek (approved):
LGTM! Thanks!

PR #52, opened 2023-04-10
Handle formatting quirks in CertificatePolicy

mprahl:
The durations returned from the Kubernetes API include zero values for proceeding time units so this makes the template-sync think the CertificatePolicy is always out of sync. The approach taken is done instead of converting the unstructured policy template to its concrete type using reflection to avoid having to depend on the certificate-policy-controller Go module and needing to stay up to date for API changes. Additionally, this would not handle the special case of pruneObjectBehavior in a ConfigurationPolicy since the default value is filled in by Kubernetes and not from the Go types being marshalled.

Though CertificatePolicy isn't officially part of the OCM community, it is a common third-party policy type so the template-sync should handle this correctly.

### DIFF
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
+	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


mprahl:
This isn't needed anymore after this PR:
https://github.com/stolostron/cert-policy-controller/pull/202

PR #54, opened 2023-04-12
Reset the complianceState when Pending

JustinKuli:
The situation this is meant to fix is not fully understood. But in case a policy template that was previously pending somehow has some status that has not yet made its way into the Policy's status, this will reset it, so that the policy controller knows to send a new compliance event.

Refs:
 - https://issues.redhat.com/browse/ACM-4699

### DIFF
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -1246,10 +1246,9 @@ func (r *PolicyReconciler) emitTemplateEvent(
 
 // handleSyncSuccess performs common actions that should be run whenever a template is in sync,
 // whether there were changes or not. If no changes occurred, an empty message should be passed in.
-// If the given policy template was in a template-error state (determined by checking the status),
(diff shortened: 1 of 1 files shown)


dhaiducek (approved):
/hold for other reviews

This is an odd corner of code, but this does make sense to resolve the E2E test. I guess resetting the compliance forces the policy controller to send another event to verify its compliance or something?

JustinKuli:
> I guess resetting the compliance forces the policy controller to send another event to verify its compliance?

Yes exactly. We already did this for some cases involving template-errors. I think it could be the case whenever the latest status was supplied by the template-sync controller, instead of the policy-controller. So far I think (hope) that the only situations for that are when it goes into Pending, and when it has a template-error.

The policy-controller doesn't really know about the Policy (how confusing is that? 😄) so it doesn't see those statuses. So it assumes the last status is the last one it sent, and it doesn't need to send another.

PR #55, opened 2023-04-17
upgrade dependents

skeeey:
1. upgrade kube to 1.26
2. upgrade controller-runtime to v0.14.5
3. upgrade addon-framework to v0.6.1


go.mod
```diff
@@ -5,108 +5,91 @@ go 1.19
...
-	sigs.k8s.io/kube-storage-version-migrator v0.0.4 // indirect
-	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
+	sigs.k8s.io/json v0.0.0-20220713155537-f223a00ba0e2 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
-replace (
-	golang.org/x/crypto => golang.org/x/crypto v0.0.0-20220722155217-630584e8d5aa // CVE-2021-43565
-	golang.org/x/text => golang.org/x/text v0.3.8 // CVE-2022-32149
```
dhaiducek, line 110: Removing these replacements re-introduced vulnerable packages presented by the referenced CVEs. Please restore them.

dhaiducek (changes requested):

go.mod
skeeey: restored

mprahl:
@dhaiducek and @skeeey we can't upgrade controller-runtime because we need to support OCP 3.11 which doesn't support leases for leader election, so we must use configmap locking which was removed in newer versions of client-go, which newer controller-runtimes require.

skeeey:
@dhaiducek and @mprahl thanks for review

Now we depend the policy in next-gen controlplane directly (https://github.com/stolostron/multicluster-controlplane/blob/main/go.mod#L28), and the next-gen controlplane needs a newer controller-runtime, so would you consider to create a feature branch for us? then next-gen controlplane can depend on this feature branch

because the next-gen controlplane needn't to support ocp311, so we can  update the dependencies in this feature branch


Thanks again

dhaiducek:
@skeeey Thanks for the context! We'll discuss this on our team. Not being able to update the `controller-runtime` has been a pain point for us also, so we'll consider paths forward to make this feasible.

PR #56, opened 2023-04-17
Synchronously emit template events

JustinKuli:
Main commit:

Previously, the controller-runtime event recorder was used for these events. Other policy controllers have moved away from that, for various reasons. In this case, if a policy went from pending to noncompliant and back to pending, the "old" pending event would be re-used by the event recorder, and only the `lastTimestamp` would be updated. In this case, if a policy controller emitted a compliance event within the same second as the Pending event, the status-sync would see it as a tie, and use the hex-encoded nanoseconds in the event name. But the event name was not updated from the original instance when the policy was pending, so the events would be ordered incorrectly.

Most error cases from this synchronous sending can be ignored because they are already error cases that would be requeued.

Refs:
 - https://issues.redhat.com/browse/ACM-4699

### DIFF
--- a/controllers/gatekeepersync/gatekeeper_constraint_sync.go
+++ b/controllers/gatekeepersync/gatekeeper_constraint_sync.go
@@ -396,7 +396,9 @@ func (r *GatekeeperConstraintReconciler) sendComplianceEvent(
 			return nil
 		}
 
-		err := r.SendEvent(ctx, constraint, owner, msg, compliance)
+		reason := utils.EventReason(constraint.GetNamespace(), constraint.GetName())
+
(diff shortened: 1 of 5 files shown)


yiraeChristineKim:
does this pr need squash?

JustinKuli:
Succeeding 30 times in a row on the targeted framework tests: [(last 3 commits)](https://github.com/JustinKuli/governance-policy-framework/pull/8).

JustinKuli:
> does this pr need squash?

I can squash it if you want, but I felt like splitting the changes into these commits made sense.

yiraeChristineKim (approved):

PR #59, opened 2023-04-20
GoSec fixes and improvements

JustinKuli:
Fixes some issues identified by GoSec, and incorporates the scan into CI here so it will fail when other issues are found.

### DIFF
--- a/.github/workflows/kind.yml
+++ b/.github/workflows/kind.yml
@@ -47,11 +47,12 @@ jobs:
       run: |
         go mod verify
 
-    - name: Verify format
+    - name: Some quality checks
       run: |
         make fmt
         git diff --exit-code
         make lint
+        make gosec-scan
 
     - name: Verify deploy/operator.yaml
       run: |
--- a/Makefile
+++ b/Makefile
@@ -177,7 +177,7 @@ gosec:
 
 .PHONY: gosec-scan
 gosec-scan: gosec
-	$(GOSEC) -fmt sonarqube -out gosec.json -no-fail -exclude-dir=.go ./...
+	$(GOSEC) -fmt sonarqube -out gosec.json -stdout -exclude-dir=.go -exclude-dir=test ./...
 
 ############################################################
 # build section
--- a/Makefile
+++ b/Makefile
@@ -337,6 +337,15 @@ e2e-stop-instrumented:
 e2e-debug:
 	@echo local controller log:
 	-cat build/_output/controller.log
+	@echo pods on hub cluster
+	-kubectl get pods -A --kubeconfig=$(HUB_CONFIG)
+	-kubectl get pods -A -o yaml --kubeconfig=$(HUB_CONFIG)
+	@echo pods on managed cluster
+	-kubectl get pods -A --kubeconfig=$(MANAGED_CONFIG)
+	-kubectl get pods -A -o yaml --kubeconfig=$(MANAGED_CONFIG)
+	@echo gatekeeper logs on managed cluster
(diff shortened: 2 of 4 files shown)


JustinKuli:
/cc @dhaiducek

I don't know if something like this gosec change would be better in a common makefile?

Makefile
```diff
@@ -177,7 +177,7 @@ gosec:
 
 .PHONY: gosec-scan
 gosec-scan: gosec
-	$(GOSEC) -fmt sonarqube -out gosec.json -no-fail -exclude-dir=.go ./...
+	$(GOSEC) -fmt sonarqube -out gosec.json -exclude-dir=.go -exclude-dir=test ./...
```
dhaiducek, line 180: I'm trying to decide whether removing `-no-fail` would affect `stolostron`, but it's probably nicer (and less complicated) to find this out in the action than from SonarCloud.

.github/workflows/kind.yml
```diff
@@ -95,6 +95,7 @@ jobs:
       run: |
         make test-coverage
         make coverage-verify
+        make gosec-scan
```
dhaiducek, line 98: Maybe this should just have this step match `stolostron`? With the change that we `tee` to the output file rather than having `-out gosec.json` in the command? (`tee` because otherwise I don't think we'll see the output?)
```
    - name: Test Coverage and Report Generation
      if: ${{ matrix.kind == 'latest' }}
      run: |
        make test-coverage | tee report_unit.json
        make coverage-verify
        make gosec-scan | tee gosec.json
```

dhaiducek (changes requested):
Looks good--thanks for the update! We should have done this a while ago. I have some comments for consideration.

Makefile
JustinKuli: I was thinking that in `stolostron`, we would not remove `-no-fail`, and we would just be alerted to issues by SonarCloud like we are now. But finding out earlier in the action might be good...

Doing it this way would let us possibly allow "low" severity issues in PRs here, but have SonarCloud track them separately. But I don't know if that's really useful.

.github/workflows/kind.yml
JustinKuli: We could add `-stdout` if we also want the report printed. I think the "normal" output of the tool is some logs, and not the report. So `tee` might make the report unreadable by other tools.

dhaiducek: Oh nice--I was wondering about other output. Then yes let's use `-stdout`. Otherwise, the scan might fail with no indication of why.

dhaiducek: Also, now that this is blocking, maybe we should move it up with the linting so that it can be addressed more quickly?

Makefile
dhaiducek: I feel like if SonarCloud is going to fail, we might as well have it fail here also?

Also I'm not sure excluding the test dir is necessary?
> gosec will ignore test files across all packages

JustinKuli: Some of our test utilities are not in `_test.go` files, so it doesn't know to ignore them. In particular, without this, it was noticing a "major" issue because of how we call `kubectl`

JustinKuli:
/hold

I found the other test issue

dhaiducek:
> /cc @dhaiducek
> 
> I don't know if something like this gosec change would be better in a common makefile?

I was going to do it, but if you have a moment, it can go here: https://github.com/stolostron/governance-policy-framework/blob/02d52d5811bfa706ec84fcd9a2f153a46ae9aa5b/build/common/Makefile.common.mk#L108

dhaiducek (approved):

PR #60, opened 2023-04-25
Stop setting the Gatekeeper constraint as compliant without audit results

mprahl:
Relates:
https://issues.redhat.com/browse/ACM-5147


controllers/gatekeepersync/gatekeeper_constraint_sync.go
```diff
@@ -252,6 +252,12 @@ func (r *GatekeeperConstraintReconciler) Reconcile(
 			continue
 		}
 
+		if !violationsSet {
+			log.V(1).Info("The constraint audit results haven't yet been posted. Doing nothing.")
```
dhaiducek, line 256: Nit: Maybe something like "Continuing" or "Skipping status update" rather than "Doing nothing"?

dhaiducek (approved):
/hold in case we want to update the log message

LGTM! Very minor comment on the log message.

controllers/gatekeepersync/gatekeeper_constraint_sync.go
mprahl: Good point. Thanks!

mprahl:
@dhaiducek this is ready for review now.

dhaiducek (approved):

PR #62, opened 2023-05-04
Check v1beta1 CRD for policy template label

JustinKuli:
On older versions of k8s where the v1 version of CRDs is not available, this method was preventing the controller from working.

Refs:
 - https://issues.redhat.com/browse/ACM-5326

### DIFF
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -1397,16 +1397,23 @@ func (r *PolicyReconciler) hasPolicyTemplateLabel(
 	}
 
 	err := r.Get(ctx, crdName, &crd)
-	if err != nil {
-		// If it wasn't found, then it wasn't in the cache and doesn't have the label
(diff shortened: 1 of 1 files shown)


dhaiducek (approved):
I double checked the addon-controller. The CRDs don't have the label, so would you please add it there also?

- v1beta1 CertPolicy: https://github.com/stolostron/governance-policy-addon-controller/blob/a9d7348d8292a227d069d2f3199acd20243a686b/pkg/addon/certpolicy/manifests/managedclusterchart/templates/policy.open-cluster-management.io_certificatepolicy_crd.yaml#L12
- v1 CertPolicy:https://github.com/stolostron/governance-policy-addon-controller/blob/a9d7348d8292a227d069d2f3199acd20243a686b/pkg/addon/certpolicy/manifests/managedclusterchart/templates/policy.open-cluster-management.io_certificatepolicy_crd.yaml#L189

PR #63, opened 2023-05-05
Upgrade: Go v1.20 and packages

dhaiducek:
Also removes legacy leader election

ref: https://issues.redhat.com/browse/ACM-5042


controllers/statussync/policy_status_sync.go
```diff
@@ -318,10 +318,9 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 			newHistory = append(newHistory, history[historyIndex].ComplianceHistory)
 
 			for j := historyIndex; j < len(history); j++ {
-				if history[historyIndex].EventName == history[j].EventName &&
-					history[historyIndex].Message == history[j].Message {
-					// same event, filter it
-				} else {
+				// Skip over duplicate statuses where the event name and message match the current status
+				if history[historyIndex].EventName != history[j].EventName &&
+					history[historyIndex].Message != history[j].Message {
```
JustinKuli, line 323: Check my logic here, sorry 😅

I don't think this is exactly equivalent to what was here before. Previously, it was `if !(A && B)`. Now it is `if (!A && !B)`. I think the confusion is that the new condition **is** equivalent to `if !(A || B)`.

In particular: if the message is the same but the name is different, then previously it would change the `historyIndex`. But now, in that same situation, it doesn't.

To be fair, I wonder if this whole thing should be updated, it's kind of hard to reason through... but that should go in a different PR.

dhaiducek: I think you're right. Right now it'll only increment the index if they **both** don't match:

| `A` | `B` | `!A` | `!B` | `!(A && B)` | `!A && !B` | `!A \|\| !B` |
| :---: | :---: | :---: | :---: | :---: | :---: | :---: |
| 0 | 0 | 1 | 1 | 1 | 1 | 1 |
| 0 | 1 | 1 | 0 | 1 | 0 | 1 |
| 1 | 0 | 0 | 1 | 1 | 0 | 1 |
| 1 | 1 | 0 | 0 | 0 | 0 | 0 |

I think `!A || !B` should be equivalent to `!(A && B)`. I'll update it (and I guess I'll be responsible and take a minute to see why the E2E didn't catch this...).

dhaiducek: Logic updated--thanks for catching that!

dhaiducek:
Sorry for the churn--CI is passing now.

JustinKuli (approved):

PR #65, opened 2023-05-17
Update event to include real status

clyang82:
for the global hub, we are going to show the policy compliance trend. It is very important to have real policy status in the events so that we can know the frequency of policy changing.

@mprahl Is the change OK? Thanks.


mprahl (approved):
This should be okay since this event is not used for any internal processes and is informational for the user.

mprahl (approved):
This should be okay since this event is not used for any internal processes and is informational for the user.

PR #69, opened 2023-06-09
Bug: Creating two policy templates with the same name succeeds

yiraeChristineKim:
Creating a policy with identical names succeeds. The ConfigurationPolicy is repeatedly overwritten by the template-sync, swapping out the two policies, and repeatedly updating the status with alternating statuses from the two policies.

For example, this policy has two ConfigurationPolicy templates both named policy-pod

Expect to fail and return a status that there is a duplicate name instead.

Ref: https://issues.redhat.com/browse/ACM-5724
Signed-off-by: Yi Rae Kim <yikim@redhat.com>

### DIFF
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -165,6 +165,19 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 		return reconcile.Result{}, nil
 	}
 
+	// Check duplicate names in configuration-policies
+	has := hasDupName(instance)
+	if has {
+		msg := "There are duplicate names in configurationpolicies, please check the policy"
+		reqLogger.Info(msg)
+
+		for tIndex := range instance.Spec.PolicyTemplates {
+			_ = r.emitTemplateError(ctx, instance, tIndex, fmt.Sprintf("[template %v]", tIndex), false, msg)
+		}
+
+		return reconcile.Result{}, nil
+	}
+
 	// Whether policy needs a finalizer to handle cleanup
 	var addFinalizer bool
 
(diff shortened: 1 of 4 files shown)


controllers/templatesync/template_sync.go
```diff
@@ -1507,3 +1526,27 @@ func finalizerCleanup(
 func (r *PolicyReconciler) setCreatedGkConstraint(b bool) {
 	r.createdGkConstraint = &b
 }
+
+// Check duplicate names in policy-templates(configurationPolicies)
+func hasDupName(pol *policiesv1.Policy) (bool, error) {
+	templates := pol.Spec.PolicyTemplates
+
+	empty := make(map[string]struct{})
```
JustinKuli, line 1534: Nit: naming this `empty` is a little confusing. My preference would be `foundNames`.

controllers/templatesync/template_sync.go
```diff
@@ -165,6 +165,25 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 		return reconcile.Result{}, nil
 	}
 
+	// Check duplicate names in configuration-policies
+	has, err := hasDupName(instance)
+	if err != nil {
+		reqLogger.Error(err, "Failed to check duplicate configurationpolicy name")
+
+		return reconcile.Result{}, nil
+	}
```
JustinKuli, line 174: I think the error flow here could cause some problems. It looks like an error can occur if a template object definition isn't valid and can't be unmarshalled. This return makes it give up forever, without putting a status on the policy...

I think the unmarshal error could be ignored in `hasDupName`, and just skip those templates. So `hasDupName` wouldn't return an error, and the logic would be a bit simpler. The same error should appear later when the templates are being considered individually, right?

JustinKuli (changes requested):
Looks good overall, the test looks good. I'm just worried about the weird error case.

controllers/templatesync/template_sync.go
yiraeChristineKim: good great catch!

PR #71, opened 2023-07-05
Stop explicitly disabling CGO

mprahl:
Addresses CVE-2023-3089

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -246,7 +246,7 @@ kind-deploy-controller:
 	kubectl apply -f deploy/operator.yaml -n $(KIND_NAMESPACE) --kubeconfig=$(MANAGED_CONFIG)
 
 .PHONY: kind-deploy-controller-dev
-kind-deploy-controller-dev: kind-deploy-controller build-images
+kind-deploy-controller-dev: kind-deploy-controller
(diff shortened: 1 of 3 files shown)


dhaiducek (approved):

.github/workflows/kind.yml
```diff
@@ -83,7 +83,6 @@ jobs:
 
     - name: Verify Deployment Configuration
       run: |
-        make build-images
```
dhaiducek, line 86: This is explicitly being called in the framework tests also, so either this needs to stay here and the `build-images` removed from the Make target in this repo OR it needs to be removed from the framework workflow:
https://github.com/stolostron/governance-policy-framework/blob/0903452ed5de2b5d99eba286f4a16ef1007ed064/.github/workflows/kind.yml#L159-L161

dhaiducek: This PR goes with the former strategy: https://github.com/stolostron/config-policy-controller/pull/531#pullrequestreview-1507628578

mprahl: Okay, I'll remove it from the Make target instead.

dhaiducek (dismissed):

dhaiducek (approved):

PR #75, opened 2023-07-21
Allow informonly configuration policies

zyjjay:
Overrides the parent policy's remediationAction when 'informonly' is provided as the remediationAction for ConfigurationPolicy

### DIFF
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -1193,10 +1193,7 @@ func generatePendingMsg(dependencyFailures []depclient.ObjectIdentifier) string 
 
 func overrideRemediationAction(instance *policiesv1.Policy, tObjectUnstructured *unstructured.Unstructured) {
 	// override RemediationAction only when it is set on parent
-	if instance.Spec.RemediationAction == "" {
-		return
-	}
-
+	// or when a policy is set to informonly
 	if tObjectUnstructured.GroupVersionKind().Group == utils.GConstraint {
 		var enforcementAction string
 
@@ -1210,20 +1207,38 @@ func overrideRemediationAction(instance *policiesv1.Policy, tObjectUnstructured 
 		}
 
 		if spec, ok := tObjectUnstructured.Object["spec"]; ok {
-			specObject, ok := spec.(map[string]interface{})
-			if ok {
+			if specObject, ok := spec.(map[string]interface{}); ok {
 				specObject["enforcementAction"] = enforcementAction
 			}
 		}
+
+		return
 	} else if tObjectUnstructured.GroupVersionKind().Group == utils.GvkConstraintTemplate.Group {
 		// Don't override anything if it's a ConstraintTemplate
 		return
-	} else {
-		if spec, ok := tObjectUnstructured.Object["spec"]; ok {
-			specObject, ok := spec.(map[string]interface{})
-			if ok {
-				specObject["remediationAction"] = string(instance.Spec.RemediationAction)
+	}
+
+	spec, ok := tObjectUnstructured.Object["spec"]
+	if !ok {
+		return
+	}
+
+	specObject, ok := spec.(map[string]interface{})
+	if !ok {
+		return
+	}
+
+	if remediationAction, ok := specObject["remediationAction"]; ok {
+		if _, ok := remediationAction.(string); ok {
+			if strings.EqualFold(specObject["remediationAction"].(string), "informonly") {
+				specObject["remediationAction"] = strings.ToLower(string(policiesv1.Inform))
+
+				return
 			}
+		}
+
+		if instance.Spec.RemediationAction != "" {
+			specObject["remediationAction"] = string(instance.Spec.RemediationAction)
 		}
 	}
 }
--- a/test/e2e/case20_informonly_test.go
+++ b/test/e2e/case20_informonly_test.go
@@ -1,0 +1,69 @@
+// Copyright (c) 2023 Red Hat, Inc.
+// Copyright Contributors to the Open Cluster Management project
+
+package e2e
+
+import (
+	"errors"
+	"os/exec"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"open-cluster-management.io/config-policy-controller/test/utils"
+)
+
+const (
+	case20PolicyName                 string = "case20-policy-informonly"
+	case20PolicyYaml                 string = "../resources/case20_policy_informonly/case20-parent-policy.yaml"
+	case20ConfigPlcName              string = "create-configmap"
+	case20ConfigPlcNoRemediationName string = "create-configmap-no-remediationaction"
+	case20PolicyNoRemediationName    string = "case20-policy-informonly-no-remediationaction"
+	case20PolicyNoRemediationYaml    string = "../resources/case20_policy_informonly/" +
+		"case20-parent-policy-noremediation.yaml"
+)
+
+func checkInformAction(cfplc string) {
+	By("Checking template policy remediationAction")
+	Eventually(func() interface{} {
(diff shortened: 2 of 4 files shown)


controllers/templatesync/template_sync.go
```diff
@@ -1222,7 +1222,11 @@ func overrideRemediationAction(instance *policiesv1.Policy, tObjectUnstructured
 		if spec, ok := tObjectUnstructured.Object["spec"]; ok {
 			specObject, ok := spec.(map[string]interface{})
 			if ok {
-				specObject["remediationAction"] = string(instance.Spec.RemediationAction)
+				if strings.EqualFold(specObject["remediationAction"].(string), "informonly") {
```
JustinKuli, line 1225: Because the policy template inside a Policy isn't strictly typed (and it can't be strictly typed because we don't know if it will be a ConfigurationPolicy, or a Gatekeeper resource, or something else entirely), we need to be careful every time we look up a field and expect it to be a certain type. In this case, if the field doesn't exist, or isn't a string, I think this code will panic and crash the controller. So I'd rather see the lookup and the string type assertion be separated, and individually checked - it's a hassle, but it helps in case there are typos introduced by the user.

test/e2e/case20_informonly_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	Describe("Override remediationAction in spec", func() {
+		Context("When the CR does not exist", func() {
+			It("Should be noncompliant", func() {
+				Eventually(func() interface{} {
+					plc := utils.GetWithTimeout(clientManagedDynamic, gvrConfigurationPolicy,
+						case20ConfigPlcName, clusterNamespace, true, defaultTimeoutSeconds)
+
+					return utils.GetComplianceState(plc)
+				}, defaultTimeoutSeconds, 1).Should(Equal("NonCompliant"))
```
JustinKuli, line 58: This test needs a running configuration-policy-controller to set the compliance state on that ConfigurationPolicy. In these tests, only the framework-addon is running, so that won't work. I think the test can just be removed - we don't need to worry about the compliance in these tests, only the `remediationAction`.

test/e2e/case20_informonly_test.go
```diff
@@ -0,0 +1,97 @@
...
+			})
+
+			It("Should have remediationAction=inform", func() {
+				Eventually(func() interface{} {
+					plc := utils.GetWithTimeout(clientManagedDynamic, gvrConfigurationPolicy,
+						case20ConfigPlcName, clusterNamespace, true, defaultTimeoutSeconds)
+
+					return plc.Object["spec"].(map[string]interface{})["remediationAction"]
+				}, defaultTimeoutSeconds, 1).Should(Equal("inform"))
+			})
```
JustinKuli, line 94: So I think this is the "real" test we're interested in. I think we should also have another test for when the Policy doesn't have a "top-level" `remediationAction` set - what happens then?

JustinKuli (changes requested):

controllers/templatesync/template_sync.go
```diff
@@ -1193,11 +1193,29 @@ func generatePendingMsg(dependencyFailures []depclient.ObjectIdentifier) string
...
-	}
+	// or when a policy is informonly
+	if tObjectUnstructured.GroupVersionKind().Group != utils.GConstraint &&
+		tObjectUnstructured.GroupVersionKind().Group != utils.GvkConstraintTemplate.Group {
+		if spec, ok := tObjectUnstructured.Object["spec"]; ok {
+			specObject, ok := spec.(map[string]interface{})
+			if ok {
+				if remediationAction, ok := specObject["remediationAction"]; ok {
+					_, ok := remediationAction.(string)
+					if ok {
```
JustinKuli, line 1204: All of the logic looks right. But it's not as easy to read as I would prefer. My personal preference is to reduce indentation when possible.

Since this is a lot of checks, one trick would be to consider what should be done when a check fails: in this case, it should just return without changing any settings. So the first two checks could be re-written like this:

```go
spec, ok := tObjectUnstructured.Object["spec"]
if !ok {
    return // template didn't have a spec
}

specObject, ok := spec.(map[string]interface{})
if !ok {
    return // template's spec wasn't a map
}
...
```

Maybe in general this function should return an error (those comments indicate what might go into the error message), but it hasn't historically, so I don't think there's a need to add that now.

JustinKuli: Alternately, you could save lines by using the `if ___; ok {` thing every time, at least to be consistent.

```suggestion
		if spec, ok := tObjectUnstructured.Object["spec"]; ok {
			if specObject, ok := spec.(map[string]interface{}); ok {
				if remediationAction, ok := specObject["remediationAction"]; ok {
					if remAction, ok := remediationAction.(string); ok {
```

test/e2e/case20_informonly_test.go
```diff
@@ -0,0 +1,71 @@
...
+	By("Checking template policy remediationAction")
+	Eventually(func() interface{} {
+		plc := utils.GetWithTimeout(clientManagedDynamic, gvrConfigurationPolicy,
+			case20ConfigPlcName, clusterNamespace, true, defaultTimeoutSeconds)
+
+		return plc.Object["spec"].(map[string]interface{})["remediationAction"]
+	}, defaultTimeoutSeconds, 1).Should(Equal("inform"))
+}
+
+var _ = Describe("Test 'InformOnly' ConfigurationPolicies", Ordered, func() {
```
JustinKuli, line 34: The `BeforeAlls` and `AfterAlls` being defined in different places make this tricky to understand.

`BeforeAll` is for setup that multiple tests will use. In this case, I don't think it's a good fit for creating the parent policy, since each test uses a different parent policy. I think the creation of that policy could just be inside the `It` for the test.

`AfterAll` is for cleanup that can happen after all the tests here run. It is good for cleaning up the policies, because it is guaranteed to run, even if a test fails. But in this case I think a better `AfterAll` would be to clean up both policies and templates used by these tests... Just be sure that if a policy or template isn't present, that doesn't cause an error.

I think if you try that, the tests could look more like:
- `AfterAll`: cleanup everything used in this test file
- `It`: 
    - create a parent policy with remediationAction=enforce in the template
    - check the template has remediationAction=inform
- `It`:
    - create a parent policy with the remediationAction unset in the template
    - check the template has remediationAction=inform

test/resources/case20_policy_informonly/case20-parent-policy.yaml
```diff
@@ -0,0 +1,27 @@
...
+    policy.open-cluster-management.io/root-policy: policy-informonly
+spec:
+  remediationAction: enforce
+  disabled: false
+  policy-templates:
+    - objectDefinition:
+        apiVersion: policy.open-cluster-management.io/v1
+        kind: ConfigurationPolicy
+        metadata:
+          name: create-configmap
```
JustinKuli, line 17: With my other suggestion around cleanup, I think the two ConfigurationPolicies would need distinct names.

JustinKuli (changes requested):
It probably seems like I'm being very picky (and that's kind of true), but I want to be sure this is as good as it can be.

controllers/templatesync/template_sync.go
zyjjay: With regards to this, I was trying to be consistent in the way the assertions were already being made in the code. Should I change all the comma-ok assertions in this function so that they are each evaluated in one line?

JustinKuli: Ah, now I see that the existing case had the one-liner for the lookup, but had it separated for the type assertion. So this was consistently following the inconsistency 😅 

Honestly, I'll accept any form you're happy with (as long as it's not much crazier than what is already in the PR). As I said, the logic looks right. De-tangling the test cases was my main reason for requesting changes on the PR, otherwise I would've approved it with these as optional suggestions.

JustinKuli (approved):
LGTM! Thanks for the adjustments!

PR #76, opened 2023-07-27
Remove duplicated compliance in Gatekeeper synced policy message

zyjjay:
ref: https://issues.redhat.com/browse/ACM-6346

### DIFF
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -1260,7 +1260,7 @@ func overrideRemediationAction(instance *policiesv1.Policy, tObjectUnstructured 
 func (r *PolicyReconciler) emitTemplateSuccess(
 	ctx context.Context, pol *policiesv1.Policy, tIndex int, tName string, clusterScoped bool, msg string,
 ) error {
-	err := r.emitTemplateEvent(ctx, pol, tIndex, tName, clusterScoped, "Normal", "Compliant; ", msg)
+	err := r.emitTemplateEvent(ctx, pol, tIndex, tName, clusterScoped, "Normal", policiesv1.Compliant, msg)
 	if err != nil {
 		tlog := log.WithValues("Policy.Namespace", pol.Namespace, "Policy.Name", pol.Name, "template", tName)
 		tlog.Error(err, "Failed to emit template success event")
@@ -1276,7 +1276,7 @@ func (r *PolicyReconciler) emitTemplateError(
 	ctx context.Context, pol *policiesv1.Policy, tIndex int, tName string, clusterScoped bool, errMsg string,
 ) error {
 	err := r.emitTemplateEvent(ctx, pol, tIndex, tName, clusterScoped,
-		"Warning", "NonCompliant; template-error; ", errMsg)
+		"Warning", policiesv1.NonCompliant, "template-error; "+errMsg)
 	if err != nil {
 		tlog := log.WithValues("Policy.Namespace", pol.Namespace, "Policy.Name", pol.Name, "template", tName)
 		tlog.Error(err, "Failed to emit template error event")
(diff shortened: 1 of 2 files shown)


dhaiducek:
@zyjjay Please add the Jira issue to the commit message so that the message propagates to `stolostron` in the commit, for example:
```
Remove duplicated compliance in Gatekeeper synced policy message

ref: https://issues.redhat.com/browse/ACM-6346
Signed-off-by: Jason Zhang <jaszhang@redhat.com>
```

controllers/utils/events.go
```diff
@@ -32,8 +32,6 @@ func (c *ComplianceEventSender) SendEvent(
 	msg string,
 	compliance policyv1.ComplianceState,
 ) error {
-	msg = string(compliance) + "; " + msg
-
```
dhaiducek, line 36: I think removing it here will remove the status from ALL messages (ConfigurationPolicy, IamPolicy, CertificatePolicy, Gatekeeper) rather than just the duplicated Gatekeeper messages.

dhaiducek: Assuming that's the case and this is affecting all policies, the cause might be somewhere in the Gatekeeper sync code instead: https://github.com/open-cluster-management-io/governance-policy-framework-addon/blob/main/controllers/gatekeepersync/gatekeeper_constraint_sync.go

zyjjay: From what I can gather, the cause seems to come from calls to the `emitTemplateEvent()` function in `template_sync.go`. The `msgMeta` field is supplied with the extra compliance text and concatenated with the compliance state in the subsequent call to `sendEvent()` in `events.go`.

JustinKuli (commented):
I agree with Dale, I think the problem is more likely in something gatekeeper-specific, otherwise I feel like we'd have noticed "NonCompliant; NonCompliant; template-error;" in other messages.

I could be convinced if there was a test for this: maybe something using this `checkForEvent` function? https://github.com/open-cluster-management-io/governance-policy-framework-addon/blob/main/test/e2e/case10_error_test.go#L311. I'm just brainstorming

controllers/utils/events.go
zyjjay: There is also a similar function in `gatekeeper_contraint_sync.go` called `sendComplianceEvent` (https://github.com/open-cluster-management-io/governance-policy-framework-addon/blob/main/controllers/gatekeepersync/gatekeeper_constraint_sync.go#L347) that wraps around `sendEvent`. In here, the `msg` field is not supplied with a compliance prefix.

dhaiducek: Ah, I see what you mean about the `SendEvent` function is appending the compliance state, so I'm thinking this isn't just limited to Gatekeeper policies and that's just a coincidence I noticed it there. 🤔 
It might be best to stick a check or two in an E2E to check these messages (if they're not checked already) since it feels like this code is arriving here from a few different places.
Also `emitTemplateError/Success/Pending` could instead pass `policiesv1.ComplianceState` directly instead of `msgMeta`, and `emitTemplateError` could add the `template-error; ` string to the `msg` instead of `msgMeta`?

zyjjay: I was actually thinking about moving the `template-error; ` string to the `msg` field but wasn't sure if I should since it seemed like a piece of  "metadata". I'll go ahead and make those changes now.

controllers/templatesync/template_sync.go
```diff
@@ -1249,7 +1249,7 @@ func overrideRemediationAction(instance *policiesv1.Policy, tObjectUnstructured
 func (r *PolicyReconciler) emitTemplateSuccess(
 	ctx context.Context, pol *policiesv1.Policy, tIndex int, tName string, clusterScoped bool, msg string,
 ) error {
-	err := r.emitTemplateEvent(ctx, pol, tIndex, tName, clusterScoped, "Normal", "Compliant; ", msg)
+	err := r.emitTemplateEvent(ctx, pol, tIndex, tName, clusterScoped, "Normal", string(policiesv1.Compliant), msg)
```
mprahl, line 1252: Instead, could you please change `emitTemplateEvent` to accept a compliance state argument instead of `msgMeta`?  It seems only in the case of `emitTemplateError` does it add an additional prefix, so this can just be added to the message instead.

controllers/templatesync/template_sync.go
```diff
@@ -1314,10 +1314,10 @@ func (r *PolicyReconciler) emitTemplatePending(
 // msg are concatenated without spaces, so any spacing should be included inside the msgMeta string.
```
mprahl, line 1314: Could you please update this comment since it still mentions `msgMeta`?

mprahl:
Since it seems to be a more general fix than just with Gatekeeper, could you please change the commit title to something like `Remove duplicated compliance in some policy compliance messages`?

test/e2e/case17_gatekeeper_sync_test.go
```diff
@@ -251,6 +251,22 @@ var _ = Describe("Test Gatekeeper ConstraintTemplate and constraint sync", Order
 			)
 		}, gkAuditFrequency*3, 1).Should(Succeed())
 
+		By("Checking if policy status history is correctly formatted")
```
mprahl, line 254: Without the code change, I would think this test would still pass because the issue seems to be in the template-sync's event emitting as opposed to the Gatekeeper sync controller. The original bug report mentioned Gatekeeper but that error actually originated from the template-sync when it was trying to create the Gatekeeper constraint from the `policy-templates` array.

I would adjust the existing test in case10 to make it more precise.

mprahl (changes requested):

PR #78, opened 2023-08-03
Reduce memory utilization of caching events

mprahl:
This removes all unnecessary fields in an Event object before it is cached. Additionally, a field selector is added to avoid caching events unrelated to policy compliance.

Relates:
https://issues.redhat.com/browse/ACM-6063


JustinKuli (approved):
LGTM

PR #79, opened 2023-08-11
ACM-3160: Invalid policy-templates array syntax leads to policy_system_errors_total incrementing

JeffeyL:
**Description of problem:** Currently, when a user provides an invalid policy template and the API call fails due to a syntax issue (for example, a ConfigurationPolicy containing a syntax error), the metric policy_system_errors_total increments instead of policy_user_errors_total. This causes false positive system level health check alerts.
**Solution:** Added simple checks for syntax errors to determine whether to increment user or system error metrics
**Ref:** https://issues.redhat.com/browse/ACM-3160

### DIFF
--- a/test/e2e/case21_syntax_error_metric_test.go
+++ b/test/e2e/case21_syntax_error_metric_test.go
@@ -1,0 +1,100 @@
+// Copyright (c) 2020 Red Hat, Inc.
+// Copyright Contributors to the Open Cluster Management project
+
+package e2e
+
+import (
+	"strconv"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	policyUtils "open-cluster-management.io/governance-policy-propagator/test/utils"
+
+	"open-cluster-management.io/governance-policy-framework-addon/test/utils"
+)
+
+var _ = Describe("Test proper metrics handling on syntax error", Ordered, func() {
+	userMetricName := "policy_user_errors_total"
+	systemMetricName := "policy_system_errors_total"
+	policyName := "case21-policy"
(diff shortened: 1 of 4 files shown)


test/e2e/case21_syntax_error_metric_test.go
```diff
@@ -0,0 +1,75 @@
...
+var _ = Describe("Test proper metrics handling on syntax error", Ordered, func() {
+	userMetricName := "policy_user_errors_total"
+	policyName := "case21-policy"
+	errPolicyFile := "../resources/case21_syntax_error_metric/case21-test-policy-err.yaml"
+	correctPolicyFile := "../resources/case21_syntax_error_metric/case21-test-policy.yaml"
+
+	cleanup := func() {
+		By("Deleting a policy on hub cluster in ns:" + clusterNamespaceOnHub)
+		// Clean up and ignore any errors (in case it was deleted previously)
+		_, _ = kubectlHub("delete", "-f", errPolicyFile, "-n", clusterNamespaceOnHub)
```
yiraeChristineKim, line 26: Adding "--ignore-not-found" is good

test/e2e/case21_syntax_error_metric_test.go
```diff
@@ -0,0 +1,75 @@
...
+		By("Checking for the " + userMetricName + " metric on the template-sync controller")
+		values := []string{}
+		Eventually(func() []string {
+			values = utils.GetMetrics(userMetricName, policyName)
+
+			return values
+		}, defaultTimeoutSeconds, 1).Should(HaveLen(1))
+		Expect(strconv.Atoi(values[0])).To(BeNumerically(">=", 1))
+	})
+})
```
yiraeChristineKim, line 75: How about policy_system_errors_total? Does this increase? I think adding the test that verify policy_system_errors_total is good

yiraeChristineKim (approved):

PR #82, opened 2023-09-01
Fix template sync duplicate detection

gparvin:
We don't want duplicate config policy templates, but the logic prevented me from having a Constraint and a ConstraintTemplate that had the same name.  Of course a config policy couldn't have that same name either so adding gvk in the name check.

Refs:
 - https://issues.redhat.com/browse/ACM-7265


yiraeChristineKim (dismissed):

yiraeChristineKim:
simple format fix needed

gparvin:
Fixed some lint issues that I forgot to check for...

yiraeChristineKim (approved):

PR #83, opened 2023-09-05
Bug: ACM-6803, Incorrect Policy does not show any validation errors and no status

JeffeyL:
Description of problem: A policy with a user error is not validated, and the policy does not become `NonCompliant` (it actually stays `Compliant`). This is because the code in the policy status sync controller completely skips of all other templates once it is unable to decode a template. If no other previously evaluated templates were `NonCompliant`, then the entire policy becomes `Compliant`.

Solution:
- In the template sync controller, template names were changed from `[template %v]` format to `template-%v` in order to fit with the regex in the policy status sync controller.
- The policy status sync controller was changed to not `break` when unable to read a template and instead to continue evaluating while setting the policy to `NonCompliant`.

Ref: https://issues.redhat.com/browse/ACM-6803

### DIFF
--- a/controllers/statussync/policy_status_sync.go
+++ b/controllers/statussync/policy_status_sync.go
@@ -215,17 +215,23 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 
 	reqLogger.Info("Updating status for policy templates")
 
-	for _, policyT := range instance.Spec.PolicyTemplates {
+	for i, policyT := range instance.Spec.PolicyTemplates {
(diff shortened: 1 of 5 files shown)


mprahl:
@JeffeyL , the commit of "Merge branch 'ACM-6803-dev' into ACM-6803" looks funky. Could you please rebase on the `main` branch and drop that commit?

JeffeyL:
Yes sorry, I changed it to WIP while I fix the commits

JeffeyL:
I didn't realize the difference between normal and hosted mode tests prior to this - I think I understand it better so I'm changing this to WIP while I fix the tests.

PR #84, opened 2023-09-15
Adjust the error to check for the ConstraintTemplate CRD missing

mprahl:
After updating to controller-runtime 0.15, it seems that a different error is returned when a CRD is missing, so it caused an API server hit every time the policy template sync controller reconciled. This uses the new error.

Relates:
https://github.com/stolostron/backlog/issues/27375

### DIFF
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -21,7 +21,6 @@ import (
 	extensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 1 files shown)


controllers/templatesync/template_sync.go
```diff
@@ -1026,7 +1026,7 @@ func (r *PolicyReconciler) cleanUpExcessTemplates(
 					})
 				}
 				// Log and ignore other errors to allow cleanup to continue since Gatekeeper may not be installed
-			} else if meta.IsNoMatchError(err) {
+			} else if discovery.IsGroupDiscoveryFailedError(errors.Unwrap(err)) {
```
dhaiducek, line 1029: Should this be updated below for `v1` also?

controllers/templatesync/template_sync.go
```diff
@@ -995,7 +995,7 @@ func (r *PolicyReconciler) cleanUpExcessTemplates(
 					namespaced: false,
 				})
 			}
-		} else if meta.IsNoMatchError(err) {
+		} else if discovery.IsGroupDiscoveryFailedError(errors.Unwrap(err)) {
```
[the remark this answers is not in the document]
JustinKuli: Or maybe it would be `errors.As` instead... error types have only actually gotten more confusing to me recently :(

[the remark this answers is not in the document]
mprahl: It was a little tricky to get to work but I got it.

controllers/templatesync/template_sync.go
mprahl: It should be there. Did GitHub now show the diff correctly for you?

dhaiducek: I don't see these lines changed in the PR:
https://github.com/open-cluster-management-io/governance-policy-framework-addon/blob/03bd852e4b4203506256317b4b9879fdbda5a105/controllers/templatesync/template_sync.go#L1062
https://github.com/open-cluster-management-io/governance-policy-framework-addon/blob/03bd852e4b4203506256317b4b9879fdbda5a105/controllers/templatesync/template_sync.go#L1430

mprahl: I see what you mean now. Fixed!

dhaiducek (approved):
LGTM!

PR #85, opened 2023-09-19
Allow duplicate naming for different kinds

gparvin:
You can't have duplicate policy templates since they will show in the console with the same name and can't be distinguished.  This PR is to allow a constraint and constraint template to have the same name and use the kind to distinguish them.

Refs:
 - https://issues.redhat.com/browse/ACM-7566

### DIFF
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -367,6 +367,13 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 			continue
 		}
 
+		// if there's a duplicate name, append the kind since duplicates should only exist for different kinds
+		for _, tmpName := range templateNames {
+			if tmpName == tName {
+				tName = fmt.Sprintf("%s-%s", tName, strings.ToLower(object.GetObjectKind().GroupVersionKind().Kind))
+			}
+		}
+
 		templateNames = append(templateNames, tName)
 
 		tLogger := reqLogger.WithValues("template", tName)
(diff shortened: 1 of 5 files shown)


controllers/templatesync/template_sync.go
```diff
@@ -367,6 +367,13 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 			continue
 		}
 
+		// if there's a duplicate name, append the kind since duplicates should only exist for different kinds
+		for _, tmpName := range templateNames {
+			if tmpName == tName {
+				tName = fmt.Sprintf("%s-%s", tName, strings.ToLower(object.GetObjectKind().GroupVersionKind().Kind))
```
mprahl, line 373: Let's use an invalid Kubernetes object name format for this. Also, I think you can just use `GetKind()`.

Also, why is the match case insensitve? Is Kubernetes not case sensitive for the `kind` field?
```suggestion
				tName = fmt.Sprintf("%s:%s", tName, strings.ToLower(object.GetKind()))
```

mprahl: Oh, I see. You're actually modifying the template name. I think we should avoid this as this is confusing behavior for the user.

Is the issue that the status sync gets confused for which template is being referenced? If that's the case, then it should refer back to the event.Related field (if present) which says which policy template that status event is for and determine which policy template that way. For the Gatekeeper controller, it always sets it:
https://github.com/open-cluster-management-io/governance-policy-framework-addon/blob/da526413602d46c949a79ef867ee57743a468024/controllers/utils/events.go#L69-L75

mprahl: If it's truly a console issue like you say in the PR description, then we should have a console fix instead.

PR #86, opened 2023-09-27
Fix RegEx to allow any character in a template name

dhaiducek:
If an invalid character is provided in a template name, our RegEx was filtering it out and not handling it properly.

I'm not sure whether this is too broad and we might need something like non-whitespace `^\s` or non-forward-slash `^/`, but technically a user can put whatever they want in there, so it seemed like `.` might be most appropriate.

ref: https://issues.redhat.com/browse/ACM-7737


dhaiducek:
The E2E tests failed because I was expecting only one history status and got 7. I've added logic here to not requeue when an error matching `k8serrors.IsInvalid()` occurs since it seems that's unrecoverable and shouldn't be tried again, but if that's too risky it's in a separate commit and can be removed.

gparvin (approved):

PR #91, opened 2023-10-11
Enable concurrent reconciles, QPS, and burst configurations

zyjjay:
To allow configuration for concurrent reconciles, QPS, and burst, the following changes are proposed:
- Each reconciler struct for all controllers will require an extra field called `ConcurrentReconciles`, which is exposed in main and set to the value specified in the arguments passed to the framework addon.
- The `ConcurrentReconcile` field will be used when the set up between controller and manager occurs
-  QPS and burst are also parsed from controller arguments passed to the framework addon, and override default QPS and burst values from controller-runtime after the hub and managed configs are created.

ref: https://issues.redhat.com/browse/ACM-6711

### DIFF
--- a/main.go
+++ b/main.go
@@ -196,6 +196,10 @@ func main() {
 		}
 	}
 
+	// Override default QPS (20) and Burst (30) to client configurations
+	managedCfg.QPS = tool.Options.ClientQPS
+	managedCfg.Burst = int(tool.Options.ClientBurst)
+
 	mgrOptionsBase := manager.Options{
 		LeaderElection: tool.Options.EnableLeaderElection,
 		// Disable the metrics endpoint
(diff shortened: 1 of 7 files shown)


main.go
```diff
@@ -196,6 +196,10 @@ func main() {
 		}
 	}
 
+	// Override default QPS (20) and Burst (30) to client configurations
+	hubCfg.QPS, hubCfg.Burst = tool.Options.ClientQPS, int(tool.Options.ClientBurst)
+	managedCfg.QPS, managedCfg.Burst = tool.Options.ClientQPS, int(tool.Options.ClientBurst)
```
JustinKuli, line 201: Using multiple assignments like this is unusual. I'd prefer each to have their own line.

I also wonder if it would be nice to be able to independently configure the QPS and Burst for the hub cluster client and the managed cluster client. I can imagine that a user would increase these settings in order to speed things up on the managed cluster, but would not want to open up the possibility of a massively increased number of requests to their hub cluster. Maybe another reviewer can provide input - we also don't want to have too many configuration options, because they could get hard to manage.

JustinKuli (approved):

PR #92, opened 2023-10-13
Update log calls to be more consistent

JustinKuli:
I was looking at this for ACM-7778, but I don't think that the logs here are particularly noisy - since the controller is event driven, knowing when it starts reconciling something (for example) can be quite useful. Instead, this PR just ensures that the controller's log messages are consistently formatted.

### DIFF
--- a/controllers/gatekeepersync/gatekeeper_constraint_sync.go
+++ b/controllers/gatekeepersync/gatekeeper_constraint_sync.go
@@ -112,7 +112,7 @@ func (r *GatekeeperConstraintReconciler) Reconcile(
 	err := r.Get(ctx, request.NamespacedName, policy)
 	if err != nil {
 		if k8serrors.IsNotFound(err) {
-			log.V(1).Info("The Policy was deleted. Cleaning up watchers and status message cache.")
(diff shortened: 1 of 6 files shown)


dhaiducek (approved):
LGTM! Thanks for the updates!

I found the `DisableStackTrace`, which might go a long way toward making logs smaller all around since I find the error logs to be particularly noisy (and somewhat frequent): https://github.com/kubernetes-sigs/controller-runtime/pull/1348

JustinKuli:
That is neat! I hadn't thought about it much. The stack traces are kind of annoying when they're attached to common update failures because the resource has already been updated or something... but I feel like they could be useful for other errors (especially when we see an error that doesn't happen very often, any extra info on how it got there could be helpful).

What's more annoying to me is that the stack traces aren't in the same line as the log, so when I grep for specific things, they become unattached.

dhaiducek:
> What's more annoying to me is that the stack traces aren't in the same line as the log, so when I grep for specific things, they become unattached.

...though that may be more because we're using `console` logging rather than `json` logging? Just a guess, though...

PR #105, opened 2024-01-05
Handle policy dependencies of Gatekeeper types

JustinKuli:
Originally, policy dependencies could only handle objects that had a `status.complianceState` field. Gatekeeper ConstraintTemplates and Constraints do not have this field, but we assign them a compliance in a similar way elsewhere. This commit allows these Gatekeeper types to be used for policy dependencies like a user might expect.

Refs:
 - https://issues.redhat.com/browse/ACM-8998

### DIFF
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -1146,13 +1146,7 @@ func (r *PolicyReconciler) processDependencies(
 	var dependencyFailures []depclient.ObjectIdentifier
 
 	for dep := range templateDeps {
-		depGvk := schema.GroupVersionKind{
-			Group:   dep.Group,
-			Version: dep.Version,
-			Kind:    dep.Kind,
-		}
-
-		rsrc, _, err := utils.GVRFromGVK(discoveryClient, depGvk)
+		rsrc, namespaced, err := utils.GVRFromGVK(discoveryClient, dep.GroupVersionKind())
 		if err != nil {
 			tLogger.Error(err, "Could not find an API mapping for the dependency", "object", dep)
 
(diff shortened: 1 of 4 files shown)


controllers/templatesync/template_sync.go
```diff
@@ -1161,24 +1155,57 @@ func (r *PolicyReconciler) processDependencies(
...
+			}
+
+			res = dClient.Resource(rsrc).Namespace(ns)
+		} else {
+			res = dClient.Resource(rsrc)
+		}
 
 		depObj, err := res.Get(ctx, dep.Name, metav1.GetOptions{})
 		if err != nil {
+			if dep.Group == utils.GvkConstraintTemplate.Group && templateDeps[dep] != "Compliant" {
```
mprahl, line 1174: Do you need to check that the error is that it wasn't found/CRD not installed?

controllers/templatesync/template_sync.go
```diff
@@ -1161,24 +1155,57 @@ func (r *PolicyReconciler) processDependencies(
...
+		switch dep.Group {
+		case utils.GvkConstraintTemplate.Group:
+			if templateDeps[dep] != "Compliant" {
+				// The ConstraintTemplate was found, but the policy wants it to not be found
+				tLogger.Info("Compliance mismatch for dependency object", "object", dep)
+
+				dependencyFailures = append(dependencyFailures, dep)
+			}
+		case utils.GConstraint:
+			violations, found, err := unstructured.NestedInt64(depObj.Object, "status", "totalViolations")
```
mprahl, line 1194: I'm thinking that a more generic way rather than duplicate the logic to determine compliance is for any policy template that isn't an OCM policy and is supported, check the compliance according to the replicated root policy. You should be able to find the root policy name by checking the Gatekeeper constraint template/constraint's `policy.open-cluster-management.io/policy` label.

JustinKuli: That sounds nice, but it shifts some complexity to managing the list of objects to watch: right now I like that the objects a policy watches is just the list of all its dependencies, exactly as they are defined.

The original design was that "dependencies will be able to accept any object that has a status.complianceState field" - technically it isn't necessary for that object to be managed by the policy framework.

controllers/templatesync/template_sync.go
JustinKuli: That kind of context could be nice, I'll see what I can do.

controllers/templatesync/template_sync.go
mprahl: This is a simple enough check that this is okay.

controllers/templatesync/template_sync.go
JustinKuli: I think I misunderstood what you were actually asking, and started implementing a way to help users understand reasons for a dependency not being satisfied... which is still good, and helpful! But I think you were actually just requesting that this case be checked to ensure that the object was really "not found" as opposed to some other error?

mprahl: That's right. It looks like you fixed that in the latest push.

controllers/templatesync/template_sync.go
```diff
@@ -1170,14 +1178,22 @@ func (r *PolicyReconciler) processDependencies(
 		}
 
 		depObj, err := res.Get(ctx, dep.Name, metav1.GetOptions{})
-		if err != nil {
+		if k8serrors.IsNotFound(err) {
```
mprahl, line 1181: I think you'll also need to check `errors.As(err, &errGroupDiscoveryFailed)` since Gatekeeper/ConstraintTemplate CRD is not guaranteed to be installed.

mprahl: Actually, I see further up you error early if the GVR can't be found.

controllers/templatesync/template_sync.go
```diff
@@ -1135,69 +1135,108 @@ func (r *PolicyReconciler) cleanUpExcessTemplates(
...
-
-				dependencyFailures = append(dependencyFailures, dep)
+				dependencyFailures[dep] = DepFailWrongCompliance
 			}
 		}
+
+		if reason, failed := dependencyFailures[dep]; failed {
+			tLogger.Info("Dependency not satisfied", "reason", reason, "object", dep)
+		} else {
+			tLogger.Info("Dependency satisfied", "object", dep)
```
mprahl, line 1227: This might be a bit verbose
```suggestion
			tLogger..V(1).Info("Dependency satisfied", "object", dep)
```

controllers/templatesync/template_sync.go
```diff
@@ -1135,69 +1135,108 @@ func (r *PolicyReconciler) cleanUpExcessTemplates(
...
+		}
 	}
 
 	return dependencyFailures
 }
 
 // generatePendingMsg formats the list of failed dependencies into a readable error.
 // Example: `Dependencies were not satisfied: 1 is still pending (FooPolicy foo)`
```
mprahl, line 1235: Does this example need to change?

controllers/templatesync/template_sync.go
```diff
@@ -1135,69 +1135,108 @@ func (r *PolicyReconciler) cleanUpExcessTemplates(
...
-func generatePendingMsg(dependencyFailures []depclient.ObjectIdentifier) string {
-	names := make([]string, len(dependencyFailures))
-	for i, dep := range dependencyFailures {
-		names[i] = fmt.Sprintf("%s %s", dep.Kind, dep.Name)
+func generatePendingMsg(dependencyFailures map[depclient.ObjectIdentifier]string) string {
+	names := make([]string, 0, len(dependencyFailures))
+	for dep := range dependencyFailures {
+		names = append(names, fmt.Sprintf("%s %s", dep.Kind, dep.Name))
```
mprahl, line 1239: Could you sort this so that the message is always consistent? Ranging over a map is not deterministic as far as I know.

mprahl (commented):
A couple of minor comments but afterwards, I'll approve it. This looks great!

controllers/templatesync/template_sync.go
JustinKuli: I kinda thought GVRFromGVK would handle that kind of thing.

controllers/templatesync/template_sync.go
JustinKuli: Good catch

controllers/templatesync/template_sync.go
JustinKuli: I thought about including the reason in this message, but I didn't like the way it looked - it became too complex in the interesting cases. So the format is actually still the same.

mprahl (approved):

PR #110, opened 2024-02-07
Fix requeue when Kind is missing

dhaiducek:
If the Kind is missing after decoding the JSON, we shouldn't be requeuing since this is not recoverable.
(Yes, the E2E YAML is from real life 🤦)

Also updates the Makefile (since we synced it without taking the time to clean up the repo one).

ref: https://issues.redhat.com/browse/ACM-9844


JustinKuli (approved):
LGTM!

PR #111, opened 2024-02-08
Record compliance events on the compliance events history API

mprahl:
If a policy controller sends a Kubernetes event with the policy.open-cluster-management.io/policy-compliance-db-id annotation set. The compliance event will be recorded on the compliance events history API.

Also, when the policy is disabled/deleted, a compliance event is recorded on the compliance events history API.

Note that to enable this functionality, the --compliance-api-url must be passed to this controller.

Relates:
https://issues.redhat.com/browse/ACM-6889


controllers/statussync/policy_status_sync.go
```diff
@@ -441,7 +490,230 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
+
+			events.AddRateLimited(ceUntyped)
+			events.Done(ceUntyped)
+
+			continue
+		}
+
+		httpRequest.Header.Set("Content-Type", "application/json")
+
+		if hubCfg.BearerToken != "" {
```
yiraeChristineKim, line 665: Send Cert and Token both?

controllers/statussync/policy_status_sync.go
```diff
@@ -441,7 +490,230 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
+			httpRequest.Header.Set("Authorization", "Bearer "+hubCfg.BearerToken)
+		}
+
+		httpResponse, err := httpClient.Do(httpRequest)
+		if err != nil {
+			log.Info("Failed to record the compliance event with the compliance API. Will requeue.")
```
yiraeChristineKim, line 671: Need check status check? if it is 500 do not add to `events.AddRateLimited`

controllers/templatesync/template_sync.go
```diff
@@ -999,6 +1016,8 @@ func (r *PolicyReconciler) cleanUpExcessTemplates(
 			}
 			// Iterate over the ConstraintTemplates to gather the Constraints on the cluster
 			for _, gkCT := range gkConstraintTemplateListv1.Items {
+				gkCT := gkCT
```
yiraeChristineKim, line 1019: Oh? what is this

controllers/statussync/policy_status_sync.go
```diff
@@ -441,7 +490,225 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
+		}
+
+		httpRequest.Header.Set("Content-Type", "application/json")
+
+		if hubCfg.BearerToken != "" {
+			httpRequest.Header.Set("Authorization", "Bearer "+hubCfg.BearerToken)
+		}
+
+		httpResponse, err := httpClient.Do(httpRequest)
+		if err != nil {
```
yiraeChristineKim, line 665: What if server DB is re-setting now then it needs time to set up DB then Do 
 need a delay time to re-queue this?

controllers/statussync/policy_status_sync.go
```diff
@@ -441,7 +490,225 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
+	} else if strings.HasPrefix(cleanMsg, "pending") {
+		return policiesv1.Pending
+	}
+
+	return policiesv1.NonCompliant
+}
+
+// ceRequestFromEvent converts a Kubernetes event relating to a policy controller's compliance event to a
+// struct representing a POST request to compliance events API. An error is returned if the event can't be converted.
+func ceRequestFromEvent(event *corev1.Event) (utils.ComplianceAPIEventRequest, error) {
```
yiraeChristineKim, line 511: Where handle `EventDetails.Metadata`

controllers/utils/utils.go
```diff
@@ -31,14 +34,72 @@ var (
...
+		if err == nil {
+			ce.ParentPolicy = &ComplianceAPIEventPolicyID{ID: int32(ppID)}
+		}
+	}
+
+	ce.Event = ComplianceAPIEvent{
+		Compliance: "Disabled",
+		Message:    msg,
+		Timestamp:  time.Now().Format(time.RFC3339Nano),
+		ReportedBy: "governance-policy-framework",
```
yiraeChristineKim, line 97: Where is Metadata?

controllers/utils/utils.go
```diff
@@ -31,14 +34,72 @@ var (
...
+
+type ComplianceAPIEventRequest struct {
+	UID          types.UID                   `json:"-"`
+	Cluster      ComplianceAPIEventCluster   `json:"cluster"`
+	Policy       ComplianceAPIEventPolicyID  `json:"policy"`
+	ParentPolicy *ComplianceAPIEventPolicyID `json:"parent_policy,omitempty"` //nolint:tagliatelle
+	Event        ComplianceAPIEvent          `json:"event"`
+}
+
+func GenerateDisabledEvent(
```
yiraeChristineKim, line 71: Is there GenerateEnableEvent?

test/e2e/case23_compliance_api_recording_test.go
```diff
@@ -0,0 +1,266 @@
...
+		generateEvent(ctx, "case23", "case23", "Halt, who goes there?", "Pending")
+		By("Generates a NonCompliant message")
+		generateEvent(ctx, "case23", "case23", "You shall not pass", "NonCompliant")
+		By("Generates a Compliant message")
+		generateEvent(ctx, "case23", "case23", "You may pass", "Compliant")
+
+		By("Waiting for the compliance API requests to come in")
+		Eventually(
+			func(g Gomega) {
+				lock.RLock()
```
yiraeChristineKim, line 195: Why this is needed?

controllers/statussync/policy_status_sync.go
```diff
@@ -441,7 +490,225 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
+		Timestamp:  event.LastTimestamp.Format(time.RFC3339Nano),
+		ReportedBy: "governance-policy-framework",
+	}
+
+	return ce, nil
+}
+
+// StartComplianceEventsSyncer will monitor the events queue and record compliance events on the compliance events
+// API. It uses either certificate or token authentication to authenticate with the compliance events API using the
+// configuration in hubCfg. Note that apiURL is the base URL to the API. It should not contain the path to the
```
gparvin, line 545: Right, we're using the API Server to talk to the hub not the default ingress.  It probably still needs to use reencrypt since the propagator should be expecting to see tls with the service CA not the api server ca.
I'm guessing there is no token auth option here anymore too.

main.go
```diff
@@ -312,6 +323,23 @@ func main() {
 		}
 	}()
 
+	if tool.Options.ComplianceAPIURL != "" {
+		wg.Add(1)
+
+		go func() {
+			err := statussync.StartComplianceEventsSyncer(
+				mgrCtx, tool.Options.ClusterNamespace, hubCfg, managedCfg, tool.Options.ComplianceAPIURL, queue,
```
gparvin, line 331: The hub certs can rotate/change.  A couple test scenarios are:
1. Make sure we run and go through a cert rotation ok.
2. Make sure we handle or at least understand how to recover if the customer changes their kube api server certs.

controllers/statussync/policy_status_sync.go
mprahl: I didn't see any harm in it as both won't be set but I'll add a condition to guard against setting both.

controllers/statussync/policy_status_sync.go
mprahl: An error here indicates a connection issue and not an error status code. In this case we want to retry. In the case of unexpected status codes, we want to continue retrying. If it's unauthorized, it's a misconfiguration that needs to be fixed and once fixed, we'll be able to send the events.

controllers/templatesync/template_sync.go
mprahl: The linter was failing because in Go < 1.22, the same memory address used for each item in the range so getting a pointer of `gkCT` without reassigning it will lead to strange results.

I'm not sure that this particular code would cause an issue though, but as I said, the linter started failing so I made it not fail.

controllers/statussync/policy_status_sync.go
mprahl: Good idea. I don't want to delay too much because I use `ShutDownWithDrain` to shutdown the queue when the process is requested to shutdown and I don't want it to take too long to drain the queue or else Kubernetes will forcefully stop the process.

controllers/statussync/policy_status_sync.go
mprahl: The `metadata` is for additional details related to the compliance event (not the metadata section of the Kubernetes event). It's not utilized right now and is there for use in the future such as recording the noncompliant diff of a Kubernetes object.

controllers/utils/utils.go
mprahl: Same comment as above.

test/e2e/case23_compliance_api_recording_test.go
mprahl: Because I don't want to access `requests` if the HTTP server is writing to the `requests` slice.

controllers/statussync/policy_status_sync.go
mprahl: Good point. I just changed it to use the system root CAs + the API server CA to account for both situations.

gparvin:
/hold
In case others want to take a look and have comments.  Otherwise remove the hold.
/lgtm

yiraeChristineKim (approved):
/unhold

PR #116, opened 2024-02-16
Use token authentication for the compliance history API

mprahl:
Related to https://github.com/open-cluster-management-io/governance-policy-addon-controller/pull/133

The addon framework's hub kubeconfig is certificate based. This does not allow authentication to the compliance history API using an OpenShift route that handles the TLS. Because of this, token based authentication is required.

This uses the hub kubeconfig's token first and if not set, falls back to getting the token from a secret in the cluster namespace on the Hub created by the policy addon controller.

Relates:
https://issues.redhat.com/browse/ACM-6889

### DIFF
--- a/controllers/statussync/policy_status_sync.go
+++ b/controllers/statussync/policy_status_sync.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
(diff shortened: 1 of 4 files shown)


controllers/statussync/policy_status_sync.go
```diff
@@ -721,6 +735,24 @@ func StartComplianceEventsSyncer(
...
+	}
+
+	saTokenSecret, err := client.CoreV1().Secrets(clusterNamespace).Get(
+		ctx, hubComplianceAPISAName, metav1.GetOptions{},
+	)
+	if err != nil {
+		return "", err
+	}
+
+	return string(saTokenSecret.Data["token"]), nil
```
yiraeChristineKim, line 753: Is there any case `saTokenSecret` is empty?

mprahl: Yeah, it could be empty in between the secret getting created and the secret getting populated with the token by Kubernetes.

I addressed this in the latest push. Good catch!

mprahl:
@gparvin and @yiraeChristineKim okay, this is ready for review again!

gparvin (approved):

PR #120, opened 2024-02-27
Record compliance events in the compliance history API with nanoseconds

mprahl:
This helps with ordering when compliance events have a timestamp that is the same down to the second.

Relates:
https://issues.redhat.com/browse/ACM-10155

### DIFF
--- a/controllers/statussync/policy_status_sync.go
+++ b/controllers/statussync/policy_status_sync.go
@@ -9,6 +9,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -21,7 +22,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
-	"k8s.io/apimachinery/pkg/api/errors"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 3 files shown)


test/e2e/case23_compliance_api_recording_test.go
```diff
@@ -198,16 +199,19 @@ var _ = Describe("Compliance API recording", Ordered, Label("compliance-events-a
 				g.Expect(requests[0].Event.Message).To(Equal("Halt, who goes there?"))
 				g.Expect(requests[0].ParentPolicy.KeyID).To(BeEquivalentTo(1))
 				g.Expect(requests[0].Policy.KeyID).To(BeEquivalentTo(3))
+				g.Expect(requests[0].Event.Timestamp.Nanosecond()).ToNot(Equal(0))
```
mprahl, line 202: It's possible that this could actually be 0, though it's very unlikely since it'd be a 1 in 999999999 chance for each event. If the reviewer is spooked about this flake happening, I can remove the check.

controllers/statussync/policy_status_sync.go
```diff
@@ -495,6 +489,19 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 	return reconcile.Result{}, nil
 }
 
+// parseTimestampFromEventName will parse the event name for a hexadecimal nanosecond timestamp as a suffix after a
+// period. This is a client-go convention that is repeated in the policy framework.
+func parseTimestampFromEventName(eventName string) (metav1.Time, error) {
```
yiraeChristineKim, line 494: Do we have an unit test for this?

controllers/statussync/policy_status_sync.go
```diff
@@ -535,10 +542,18 @@ func ceRequestFromEvent(event *corev1.Event) (utils.ComplianceAPIEventRequest, e
 
 	compliance := parseComplianceFromMessage(event.Message)
 
+	var timestamp metav1.Time
```
yiraeChristineKim, line 545: It would be great to have unit test for this func too

yiraeChristineKim: nono maybe optional

test/e2e/case23_compliance_api_recording_test.go
yiraeChristineKim: You mean there is a possibility of flaky test?

mprahl: Yes because it's possible the the actual nanosecond of when the event was generated was 0. I'll just change it to be a unit test instead.

yiraeChristineKim (approved):

PR #123, opened 2024-03-01
Filter Events for compliance status by Policy UID

dhaiducek:
Previously if a Policy is deleted and recreated, the controller lists all events and matches any events with the same name to the Policy, picking up events from the previously deleted Policy, which may or may not be the same template(s). This adds an additional filter to only consider Events containing the same UID.

Modifying the current E2E tests seems to cover this new case because the history lengths had to be modified since the old events were no longer being picked up since policies were being deleted between tests and recreated with the same name.

(There are also some other code update commits that I think could be backported with this change, but I can take them out if they are noisy.)

ref: https://issues.redhat.com/browse/ACM-9314


controllers/statussync/policy_status_sync.go
```diff
@@ -224,8 +224,12 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
 		// sample event.Reason -- reason: 'policy: calamari/policy-grc-rbactest-example'
 		reason := rgx.FindString(event.Reason)
-		if event.InvolvedObject.Kind == policiesv1.Kind && event.InvolvedObject.APIVersion == policiesv1APIVersion &&
-			event.InvolvedObject.Name == instance.GetName() && reason != "" {
+		// Only handle events that match the name and UID of the current Policy
+		if event.InvolvedObject.Kind == policiesv1.Kind &&
+			event.InvolvedObject.APIVersion == policiesv1APIVersion &&
+			event.InvolvedObject.Name == instance.GetName() &&
+			event.InvolvedObject.UID == instance.UID &&
+			reason != "" {
```
dhaiducek, line 232: This is the relevant change.

dhaiducek:
Sorry, I missed my E2E fiddling caused the E2E to fail! CI is passing now.

controllers/statussync/policy_status_sync.go
```diff
@@ -224,8 +224,12 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 		event := event
 		// sample event.Reason -- reason: 'policy: calamari/policy-grc-rbactest-example'
 		reason := rgx.FindString(event.Reason)
-		if event.InvolvedObject.Kind == policiesv1.Kind && event.InvolvedObject.APIVersion == policiesv1APIVersion &&
-			event.InvolvedObject.Name == instance.GetName() && reason != "" {
+		// Only handle events that match the name and UID of the current Policy
+		if event.InvolvedObject.Kind == policiesv1.Kind &&
```
mprahl, line 228: Since a UID is unique, you could avoid all these other conditions to check but it doesn't hurt.

test/e2e/case2_status_sync_test.go
```diff
@@ -72,7 +72,7 @@ var _ = Describe("Test status sync", func() {
 		}, defaultTimeoutSeconds, 1).Should(Equal(managedPlc.Object["status"]))
 	})
 	It("Should set status to NonCompliant", func() {
-		By("Generating an compliant event on the policy")
+		By("Generating an noncompliant event on the policy")
```
mprahl, line 75:
```suggestion
		By("Generating a noncompliant event on the policy")
```

controllers/statussync/policy_status_sync.go
dhaiducek: > Since a UID is unique, you could avoid all these other conditions to check but it doesn't hurt.

Haha, good point--I was just blindly adding to the list! 😂 
I'll clean it up.

mprahl (approved):
/hold for optional comments. Feel free to unhold.

mprahl (approved):
/unhold

PR #127, opened 2024-03-21
Fix template sync set to Nomcompliant when context cancel

yiraeChristineKim:
Bug: If the template-sync is shutting down and it's in the middle of a reconcile request, it can set the policy to noncompliant with this
Ref: https://issues.redhat.com/browse/ACM-10402

### DIFF
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -641,6 +641,11 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 
 			// a different error getting template object from cluster
 			resultError = err
+
+			if errors.Is(err, context.Canceled) {
+				return reconcile.Result{}, fmt.Errorf("failed to get the object in the policy template: %w", err)
+			}
+
 			errMsg := fmt.Sprintf("Failed to get the object in the policy template: %s", err)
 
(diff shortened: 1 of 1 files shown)


controllers/templatesync/template_sync_test.go
```diff
@@ -213,3 +223,143 @@ func TestHasDuplicateNames(t *testing.T) {
 		t.Fatal("Duplicate name not detected")
 	}
 }
+
+func TestContextCancel(t *testing.T) {
```
yiraeChristineKim, line 227: A Simple workaround but Not a simple test. Do we need this?

mprahl: Are you asking if we need the test?

yiraeChristineKim: Yes

mprahl: I think the test is a bit much for this, so I'm okay with removing it.

yiraeChristineKim: This test cannot catch`eObject, err := res.Get(ctx, tName, metav1.GetOptions{})`  line 534 actually

yiraeChristineKim: But I used this test to verify... in local I can control the cancel ctx point

mprahl (approved):

PR #130, opened 2024-04-10
Start and stop the Gatekeeper status sync controller dynamically

mprahl:
Rather than rely on a health probe failing if the Gatekeeper installation status changes, just dynamically start and stop a new manager.

Relates:
https://issues.redhat.com/browse/ACM-10966


main.go
```diff
@@ -951,3 +903,252 @@ func addControllers(
...
+		return err
+	}
+
+	// Disable the metrics endpoint for this manager. Note that since they both use the global
+	// metrics registry, metrics for this manager are still exposed by the other manager.
+	mgrOptions.Metrics.BindAddress = "0"
+	mgrOptions.LeaderElectionID = "governance-policy-framework-addon3.open-cluster-management.io"
+	mgrOptions.HealthProbeBindAddress = healthAddress
+
+	startMgr := func(ctx context.Context) {
```
JustinKuli, line 960: I think this could be defined outside of the `gatekeeperInstallationChecker` function. It might need more parameters, but I think it would help readability a lot.

main.go
```diff
@@ -951,3 +903,252 @@ func addControllers(
...
+	}
+
+	go func() {
+		for {
+			select {
+			case <-w.Done():
+				select {
+				case <-ctx.Done():
+					// Stop the retry watcher if the parent context is canceled.
+					w.Stop()
```
JustinKuli, line 1104: Since this is inside the `<- w.Done()` case, isn't the watcher already stopped?

main.go
```diff
@@ -951,3 +900,259 @@ func addControllers(
...
+
+	mgrRunning := false
+
+	if gatekeeperInstalled {
+		mgrRunning = true
+
+		mgrCtx, mgrCtxCancel = context.WithCancel(ctx)
+
+		go startMgr(mgrCtx)
+	}
```
JustinKuli, line 1095: Is it possible to adjust the case in the select below to do this "initial" start there? My feeling is that "recover from stop" should be the same as "initial start".

main.go
```diff
@@ -951,3 +900,262 @@ func addControllers(
 		os.Exit(1)
 	}
 }
+
+// gatekeeperInstallationChecker starts or stops the gatekeeper-constraint-status-sync controller if Gatekeeper's
+// installation status changes. The controller must be off when Gatekeeper is not installed.
+func gatekeeperInstallationChecker(
+	ctx context.Context, managedCfg *rest.Config, dynamicClient dynamic.Interface, mgrOptions manager.Options,
+) error {
```
JustinKuli, line 908: I think the cases where this function returns an error are just:
1. if it can't get a free local address
2. if the initial attempt to make the retry watcher fails.

So, if case 1 moves out of the function, and case 2 could be handled like non-initial retry watcher creation failures are handled (by retrying, I think?) then this function would not return an error.

The reason I'm thinking in that direction is that this function ends with a `go func() ...`, so I feel like most of its "work" is asynchronous. But the name of the function, and the calling convention doesn't make that obvious. If we called it as `go gkInstallChecker(...)`, it might be more clear what it is doing? Plus it would save some indents at the end.

I don't know - maybe it wouldn't be better. Just something to consider

main.go
```diff
@@ -275,12 +281,13 @@ func main() {
 		os.Exit(1)
 	}
 
-	healthAddresses := []string{mgrHealthAddr}
+	healthAddressesLock.RLock()
+	healthAddresses[mgrHealthAddr] = true
```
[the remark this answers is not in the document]
mprahl: Good catch. Thanks!

main.go
mprahl: I'm not certain about this. I couldn't find any handling of a closed context in the client-go code but it's not harmful to include this since it explicitly tells the retry loop to stop.

mprahl:
@JustinKuli could you please take another look?

main.go
```diff
@@ -291,11 +298,36 @@ func main() {
...
+			log.Error(err, "unable to detect the managed cluster's Kubernetes version")
+			os.Exit(1)
+		}
+
+		// Gatekeeper is not supported on such old versions of Kubernetes, so just always disable it in this case. In
+		// particular, CRD v1 is not available before Kubernetes v1.16.0.
+		if semver.Compare(serverVersion.GitVersion, "v1.16.0") >= 0 {
+			dynamicClient := dynamic.NewForConfigOrDie(managedCfg)
+
+			go manageGatekeeperSyncManager(mgrCtx, managedCfg, dynamicClient, mgrOptionsBase)
```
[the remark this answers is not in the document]
mprahl: Yeah, I think so!

mprahl:
> Maybe it's just had more time to settle in my head, but I think this is looking great

Thanks @JustinKuli! I addressed your comments and I had forgotten about the use of the wait group in the main function, so I took that into account when the Gatekeeper sync manager is running.

JustinKuli (approved):
LGTM

PR #132, opened 2024-04-15
Move the disabled events recording outside of spec-sync

mprahl:
Note, this is the related PR in the propagator: https://github.com/open-cluster-management-io/governance-policy-propagator/pull/215

There were issues with duplicated compliance events and having a different code path in the governance-policy-propagator for a self-managed hub. This will unifies the approaches.

It takes the approach of using a retry watcher and just generating the disabled event when the policy is deleted. The controller-runtime library cannot be used because when there is a reconcile event for a deleted object, you don't have access to the object as it was right before deletion.

Relates:
https://issues.redhat.com/browse/ACM-10418

### DIFF
--- a/controllers/complianceeventssync/disabled_events_sync.go
+++ b/controllers/complianceeventssync/disabled_events_sync.go
@@ -1,0 +1,140 @@
+package complianceeventssync
+
+import (
+	"context"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


dhaiducek (dismissed):
/hold LGTM! Holding for other reviews (or if merge coordination is needed).

mprahl:
@dhaiducek thanks for the review Dale! After thinking about this, I think I need to guarantee the watch starts from a point before the spec-sync starts in order to not miss and delete policy events. I'll address that in a bit.

/hold

JustinKuli (approved):

PR #133, opened 2024-04-18
Use the admin kubeconfig for the e2e-debug Make target

mprahl:
This fixes the debug log collection when an E2E test fails in the GitHub workflow.

Relates:
https://issues.redhat.com/browse/ACM-11200


dhaiducek (approved):
Good catch! Thanks!

PR #134, opened 2024-04-18
Use the dependency watcher cache in gatekeeper-constraint-status-sync

mprahl:
The first commit updates the E2E environment Gatekeeper to 3.15 and uses the Stolostron fork.

This also improves the situation when the ConstraintTemplate is removed and the policy still deploys a Constraint.

Relates:
https://issues.redhat.com/browse/ACM-11200

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -230,7 +230,8 @@ install-resources:
 	-kubectl apply -k deploy/hubpermissions --kubeconfig=$(HUB_CONFIG)_e2e
 	@if [ "$(KIND_VERSION)" != "minimum" ]; then \
 		echo installing Gatekeeper on the managed cluster; \
-		kubectl apply -f https://raw.githubusercontent.com/open-policy-agent/gatekeeper/v3.11.0/deploy/gatekeeper.yaml  --kubeconfig=$(MANAGED_CONFIG)_e2e; \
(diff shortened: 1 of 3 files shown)


JustinKuli (approved):
LGTM!

Makefile
```diff
@@ -230,7 +230,8 @@ install-resources:
 	-kubectl apply -k deploy/hubpermissions --kubeconfig=$(HUB_CONFIG)_e2e
 	@if [ "$(KIND_VERSION)" != "minimum" ]; then \
 		echo installing Gatekeeper on the managed cluster; \
-		kubectl apply -f https://raw.githubusercontent.com/open-policy-agent/gatekeeper/v3.11.0/deploy/gatekeeper.yaml  --kubeconfig=$(MANAGED_CONFIG)_e2e; \
+		curl -L https://raw.githubusercontent.com/stolostron/gatekeeper/release-3.15/deploy/gatekeeper.yaml | sed 's/- --disable-cert-rotation/- --disable-cert-rotation\n        - --audit-interval=10/g' | kubectl apply --kubeconfig=$(MANAGED_CONFIG)_e2e -f -; \
```
yiraeChristineKim, line 233: maybe it is better that release-3.15 being a parameter. but already merged sorr y for late

mprahl: @yiraeChristineKim won't we hit API rate limiting that way though?

PR #136, opened 2024-04-23
Handle template errors better

JustinKuli:
It appears that "special" compliance events related to template-errors can potentially be superseded by later events coming from a policy controller. Specifically, if a template is valid and gets created, but is later edited to have an error, the original template is not removed. Because undesired side-effects could be caused by removing a template (eg PruneObjectBehavior in a ConfigurationPolicy), it would be risky to "fix" this by removing the template in this situation. Instead, this PR proposes that the template-sync always re-reconcile after the status is updated, which allows it to correct the "late" event by emitting a new template-error.

The increased number of reconciles does increase the possibility of duplicate compliance events. Accordingly, some tests have been relaxed, and the duplicate detection process now refreshes the policy to try and decrease the probability. It also fixes a bug that would cause a large number of duplicated events in this scheme, because the event was being recorded on the wrong item in the history.

This PR also handles hub-template errors here, instead of in individual policy controllers.

Refs:
 - https://issues.redhat.com/browse/ACM-10858

### DIFF
--- a/controllers/templatesync/predicate.go
+++ b/controllers/templatesync/predicate.go
@@ -4,6 +4,8 @@
 package templatesync
 
 import (
+	"strings"
+
 	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
@@ -25,7 +27,29 @@ func templatePredicates() predicate.Funcs {
 			if hasAnyDependencies(updatedPolicy) {
 				// if it has dependencies, and it's not currently Pending, then
 				// it needs to re-calculate if it *should* be Pending.
-				return updatedPolicy.Status.ComplianceState != "Pending"
+				if updatedPolicy.Status.ComplianceState != "Pending" {
+					return true
+				}
+			}
+
+			// Look through the history for `template-error` events. If one is
+			// found and the "current" (0th) event is not a `template-error`, then
+			// reconcile to determine if the template-error event needs to be
+			// re-sent.
+			for _, dpt := range updatedPolicy.Status.Details {
+				if dpt == nil {
+					continue
+				}
+
+				for i, historyEvent := range dpt.History {
+					if strings.Contains(historyEvent.Message, "template-error") {
+						if i == 0 {
+							break
+						}
+
+						return true
+					}
+				}
 			}
 
 			return false
--- a/controllers/statussync/policy_status_sync.go
+++ b/controllers/statussync/policy_status_sync.go
@@ -205,7 +205,7 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 		return reconcile.Result{}, err
 	}
 	// filter events to current policy instance and build map
-	eventForPolicyMap := make(map[string]*[]historyEvent)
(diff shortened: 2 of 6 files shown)


controllers/templatesync/predicate.go
```diff
@@ -1,50 +0,0 @@
...
-	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
-	"sigs.k8s.io/controller-runtime/pkg/event"
-	"sigs.k8s.io/controller-runtime/pkg/predicate"
-)
-
-// templatePredicates filters out changes to policies that don't need to be
-// considered by the template-sync controller.
-func templatePredicates() predicate.Funcs {
-	return predicate.Funcs{
-		UpdateFunc: func(e event.UpdateEvent) bool {
```
mprahl, line 16: Rather than delete this predicate, could you check to see if the `oldPolicy` has a `template-error` compliance event and if so, always queue up the reconcile request?

mprahl:
@JustinKuli a bit of an odd thought, but in a template error situation where the current configuration policy has pruneObjectBehavior, could the template-sync disable `pruneObjectBehavior` and then delete it so that a template error situation always results in the policy template being deleted?

JustinKuli:
> @JustinKuli a bit of an odd thought, but in a template error situation where the current configuration policy has pruneObjectBehavior, could the template-sync disable `pruneObjectBehavior` and then delete it so that a template error situation always results in the policy template being deleted?

@mprahl it could... It would potentially lose the "createdByPolicy" information (which I think we already document as unreliable), and my instinct is that the process would be unreliable.

controllers/templatesync/predicate.go
JustinKuli: I suppose

mprahl:
> > @JustinKuli a bit of an odd thought, but in a template error situation where the current configuration policy has pruneObjectBehavior, could the template-sync disable `pruneObjectBehavior` and then delete it so that a template error situation always results in the policy template being deleted?
> 
> @mprahl it could... It would potentially lose the "createdByPolicy" information (which I think we already document as unreliable), and my instinct is that the process would be unreliable.

What would be unreliable? I think an update and then delete would be reliable since the config-policy-controller won't notice the deletion until it has refreshed its cache and thus seen the update.

I'm thinking that it's probably the more expected behavior to remove the previous template rather than having an in between state where old clusters have the old version and new clusters don't have a version of the policy template.

JustinKuli:
@mprahl I would prefer to more thoroughly discuss separately whether (and how) to remove templates during an error case, rather than hold this PR for the decision. I think the behavior of this PR is a step forward.

controllers/templatesync/predicate.go
```diff
@@ -25,7 +27,29 @@ func templatePredicates() predicate.Funcs {
...
+			// found and the "current" (0th) event is not a `template-error`, then
+			// reconcile to determine if the template-error event needs to be
+			// re-sent.
+			for _, dpt := range updatedPolicy.Status.Details {
+				if dpt == nil {
+					continue
+				}
+
+				for i, historyEvent := range dpt.History {
+					if strings.Contains(historyEvent.Message, "template-error") {
```
mprahl, line 45:
```suggestion
					if strings.HasPrefix(historyEvent.Message, "template-error;") {
```

PR #139, opened 2024-05-09
Report Pending status despite hub template error

JustinKuli:
[A previous change](https://github.com/open-cluster-management-io/governance-policy-framework-addon/pull/136) allowing hub-templates for all policy types, and handling any hub-template errors in this controller, caused a change in behavior. Before that change, when a policy both had a hub-template error *and* an unsatisfied dependency, the policy would be marked as Pending. The change made the hub-template error have priority.

This commit gives the Pending status priority, and adds a test for that specific behavior.

Refs:
 - https://issues.redhat.com/browse/ACM-11530


controllers/templatesync/template_sync.go
```diff
@@ -676,6 +678,17 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 			continue
 		}
 
+		// check for hub template error
+		if errAnno := metaObj.GetAnnotations()[hubTmplErrorKey]; errAnno != "" {
```
mprahl, line 682: If I'm reading this correctly, the ConfigurationPolicy that already exists would remain on the cluster even if there was a hub template error. This was also previously true in previous OCM versions, however, the config-policy-controller would not process these invalid ConfigurationPolicies. So I think this may be an unintended change in behavior.

JustinKuli: You're right. And your comment exposes the bug in the implementation just before this commit: while it prevents creation when there is a hub-template error, it doesn't remove an existing instance.

So it seems like it should delete the existing object in this situation. But now I'm wondering if we need to be very careful with configuration policies here, because of PruneObjectBehavior? What do you think

mprahl: @JustinKuli good point. It's hacky, but we could unset `pruneObjectBehavior` before deleting it. You lose the metadata on the status but at least it doesn't delete the object.

Alternatively, we could have template sync not handle ConfigurationPolicy hub template errors but handle all other policy kinds. Then the config-policy-controller would continue to handle them like before.

controllers/templatesync/template_sync.go
```diff
@@ -522,26 +513,37 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
 		if err != nil {
-			if len(dependencyFailures) > 0 {
-				// template must be pending, do not create it
-				emitErr := r.emitTemplatePending(ctx, instance, tIndex, tName, isClusterScoped,
-					generatePendingMsg(dependencyFailures))
-				if emitErr != nil {
-					resultError = emitErr
+			// not found should consider creating it
+			if k8serrors.IsNotFound(err) {
+				if len(dependencyFailures) > 0 {
```
JustinKuli, line 518: This section was moved inside the `IsNotFound` part: previously, it was possible that other errors would not be re-queued if the template was supposed to be Pending. Since those errors are rare, and this is already a complicated situation, I'm not sure if this ever caused any strange problems, but it's possible.

controllers/templatesync/template_sync.go
JustinKuli: Newly pushed change deals with it in the template-sync, and specially sets `pruneObjectBehavior` on ConfigurationPolicies in this situation.

mprahl (approved):

PR #140, opened 2024-05-22
Set the parent policy to NonCompliant when there is a duplicate name

mprahl:
The noncompliant event was sent in this case but the status-sync did not pick it up because it specified the policy template name using the array index.

Relates:
https://issues.redhat.com/browse/ACM-11806

### DIFF
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -177,13 +177,13 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 	}
 
 	// Check duplicate names in configuration-policies
-	has := hasDupName(instance)
-	if has {
-		msg := "There are duplicate names in configurationpolicies, please check the policy"
+	dupName := getDupName(instance)
+	if dupName != "" {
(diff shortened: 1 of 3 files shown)


controllers/templatesync/template_sync.go
```diff
@@ -1693,27 +1693,27 @@ func (r *PolicyReconciler) setCreatedGkConstraint(b bool) {
...
 			// Skip unmarshal error here, template error should appear later
-			return false
+			return ""
 		}
 
 		name := unstructured.GetName()
 
-		if _, has := foundNames[name]; has {
-			return true
+		if has := foundNames[name]; has {
```
dhaiducek, line 1711: Shouldn't this check the value here rather than assigning it?
```suggestion
		if foundNames[name] {
```

dhaiducek (changes requested):

controllers/templatesync/template_sync.go
mprahl: It does the same thing since `has` is the boolean value but your suggestion is more concise.

dhaiducek (approved):

PR #147, opened 2024-06-26
Set the cluster namespace on OperatorPolicy dependencies

mprahl:
Without this, the governance-policy-framework panics because it doesn't have permission to view OperatorPolicy in all namespaces.

### DIFF
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -241,15 +241,8 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 			Group:     dep.GroupVersionKind().Group,
 			Version:   dep.GroupVersionKind().Version,
 			Kind:      dep.GroupVersionKind().Kind,
-			Namespace: dep.Namespace,
+			Namespace: getDepNamespace(r.ClusterNamespace, dep),
 			Name:      dep.Name,
-		}
-
-		// Use cluster namespace for known policy types when the namespace is blank
-		if depID.Namespace == "" && depID.Group == policiesv1.GroupVersion.Group &&
-			depID.Version == policiesv1.GroupVersion.Version &&
-			strings.HasSuffix(depID.Kind, "Policy") {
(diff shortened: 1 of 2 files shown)


controllers/templatesync/template_sync.go
```diff
@@ -241,17 +241,10 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
...
 			Version:   dep.GroupVersionKind().Version,
 			Kind:      dep.GroupVersionKind().Kind,
-			Namespace: dep.Namespace,
+			Namespace: getDepNamespace(r.ClusterNamespace, dep),
 			Name:      dep.Name,
 		}
 
-		// Use cluster namespace for known policy types when the namespace is blank
-		if depID.Namespace == "" && depID.Group == policiesv1.GroupVersion.Group &&
-			depID.Version == policiesv1.GroupVersion.Version &&
```
mprahl, line 250: This version check was the culprit since OperatorPolicy is v1beta1 and not v1.

controllers/templatesync/template_sync.go
```diff
@@ -1310,13 +1308,7 @@ func (r *PolicyReconciler) processDependencies(
...
 
 		if namespaced {
-			ns := dep.Namespace
-			if ns == "" && dep.Group == policiesv1.GroupVersion.Group {
-				// ocm policies should always be in the cluster namespace
-				ns = r.ClusterNamespace
-			}
-
-			res = dClient.Resource(rsrc).Namespace(ns)
+			res = dClient.Resource(rsrc).Namespace(dep.Namespace)
```
dhaiducek, line 1311: Should this also use the new function?
```suggestion
			res = dClient.Resource(rsrc).Namespace(getDepNamespace(r.ClusterNamespace, dep))
```

mprahl: @dhaiducek it's not needed in this case because the input to `processDependencies` already has dependencies with the correct namespace.

dhaiducek (approved):

PR #148, opened 2024-07-01
Trigger status-sync reconciles when spec-sync sees mismatching status

mprahl:
Since status-sync watches the managed cluster's replicated policies on the managed cluster (not the hub), it has a blind spot for when the hub's status doesn't match and nothing triggers a status-sync reconcile (e.g. new compliance event or policy update). Since spec-sync is already watching the managed cluster's replicated policies on the hub, it can do the work of detecting the status mismatch on the hub and managed cluster, and trigger a status-sync reconcile to have the status-sync controller fix it.

Relates:
https://issues.redhat.com/browse/ACM-12447


test/e2e/case2_status_sync_test.go
```diff
@@ -117,7 +119,7 @@ var _ = Describe("Test status sync", func() {
 			return hubPlc.Object["status"]
 		}, defaultTimeoutSeconds, 1).Should(Equal(managedPlc.Object["status"]))
 	})
-	It("Should set status to Compliant again", func() {
+	It("Should set status to Compliant again", func(ctx context.Context) {
```
dhaiducek, line 122: (Optional) Use Ginkgo's `SpecContext`
ref: https://onsi.github.io/ginkgo/#interruptible-nodes-and-speccontext
```suggestion
	It("Should set status to Compliant again", func(ctx SpecContext) {
```

dhaiducek (approved):
/hold for minor E2E update and other reviews

Seems right to me! Thanks for the update!

PR #156, opened 2024-12-05
Detach ConstraintTemplate CRD log from a policy

dhaiducek:
It doesn't make sense to have a "missing CRD" message attached to any particular reconcile. For example, the log currently appears as this, which is confusing since the policy isn't a Gatekeeper policy:
```
2024-11-26T23:57:44.964Z
info
policy-template-sync
templatesync/template_sync.go:1157
The ConstraintTemplate CRD is not installed
{"Request.Namespace": "<namespace>", "Request.Name": "<policy-namespace>.<policy-name>"}
```


dhaiducek:
> It should just be the first reconcile where the controller noticed the CRD was not installed, right? Or does this get repeatedly logged (which might indicate a bug)?

Right, it's just the first reconcile. But since the CRD isn't installed, it's pretty well guaranteed to not be a Gatekeeper policy, so it's confusing to have it attached to any particular reconcile.

PR #159, opened 2025-02-07
Override the namespace set in the template

JustinKuli:
Previously, if the user set a namespace in the template, then when the template-sync tried to create the template, it would fail. Now the template-sync always uses the cluster namespace, overriding what the user might have provided.

Refs:
 - https://issues.redhat.com/browse/ACM-17666


dhaiducek (approved):
I don't particularly like that we're no longer providing feedback about the namespace to the user, but in this case since the namespace is irrelevant, I think this approach is fine. Plus, a similar approach was taken in the generator:
- https://github.com/open-cluster-management-io/policy-generator-plugin/pull/171/files

PR #160, opened 2025-02-19
Fix ConstraintTemplate creation with invalid rego

yiraeChristineKim:
ConstraintTemplate is created and compliant even with invalid rego syntax

Ref: https://issues.redhat.com/browse/ACM-15588

### DIFF
--- a/controllers/templatesync/template_sync.go
+++ b/controllers/templatesync/template_sync.go
@@ -564,6 +564,20 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 					continue
 				}
 
+				// Applicable for Gatekeeper versions v3.17 and later.
+				if isGkConstraintTemplate {
+					sentMsg, err := r.emitGKConstraintTemplateErrMsg(ctx, tObjectUnstructured,
+						res, instance, tIndex, tName)
+					if err != nil {
+						return reconcile.Result{}, err
+					}
+
+					if sentMsg {
+						// Skip success event emitting
+						continue
+					}
+				}
+
 				successMsg := fmt.Sprintf("Policy template %s created successfully", tName)
 
 				tLogger.Info("Policy template created successfully")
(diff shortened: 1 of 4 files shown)


controllers/templatesync/template_sync.go
```diff
@@ -1740,3 +1754,41 @@ func getDupName(pol *policiesv1.Policy) string {
...
+	if err != nil {
+		return false, err
+	}
+
+	created, ok, err := unstructured.NestedBool(template.Object, "status", "created")
+	if !ok || err != nil {
+		time.Sleep(500 * time.Millisecond)
+
+		return getConstraintTemplateStatus(ctx, name, res)
+	}
```
JustinKuli, line 1773: If gatekeeper isn't running, then this will recursively retry forever (or until it crashes), waiting for the status to be populated. That could prevent any other work from getting done.

Maybe it should return an error, which gets surfaced as a template-error in the policy, and cause the Reconcile to be re-queued to wait for the status to be populated.

JustinKuli (changes requested):

controllers/templatesync/template_sync.go
yiraeChristineKim: If gatekeeper is not running, ConstraintTemplate will be failed in `rec.get`

controllers/templatesync/template_sync.go
```diff
@@ -1740,3 +1759,39 @@ func getDupName(pol *policiesv1.Policy) string {
...
+) (errMsgGKT string, err error) {
+	name := tObjectUnstructured.GetName()
+
+	created, err := getConstraintTemplateStatus(ctx, name, res)
+	if err != nil {
+		return "", err
+	}
+
+	if !created {
+		return fmt.Sprintf("Failed to create Gatekeeper ConstraintTemplate. Check the status of %s.", name), nil
```
dhaiducek, line 1793: Could this happen if it just hasn't been created yet? In that case a "failed" message would be misleading. Would it be straightforward for us to pull the message out to deliver to the status?

yiraeChristineKim: There is several messages when rego is failed, My case, all of them were identical but I wasn't sure so I lead to ConstraintTemplate

controllers/templatesync/template_sync.go
yiraeChristineKim: I fixed

test/e2e/case17_gatekeeper_sync_test.go
```diff
@@ -224,7 +224,7 @@ var _ = Describe("Test Gatekeeper ConstraintTemplate and constraint sync", Order
 			}
 
 			return compliance
-		}, defaultTimeoutSeconds, 1).Should(Equal("Compliant"))
+		}, defaultTimeoutSeconds*2, 1).Should(Equal("Compliant"))
```
yiraeChristineKim, line 227: unreliable result

yiraeChristineKim:
Iine 915 add constraintTemplate as a watcher

controllers/templatesync/template_sync.go
```diff
@@ -564,6 +564,20 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 					continue
 				}
 
+				// Applicable for Gatekeeper versions v3.17 and later.
+				if isGkConstraintTemplate {
+					sentMsg, err := r.emitGKConstraintTemplateErrMsg(ctx, tObjectUnstructured,
+						res, instance, tIndex, tName)
+					if err != nil {
+						return reconcile.Result{}, err
```
yiraeChristineKim, line 572: Should we just print error message and wait for next reconcile(wait constraintTemplate changed)

JustinKuli (approved):
LGTM!

PR #164, opened 2025-06-03
Relax equivalence on template metadata

JustinKuli:
Previously, the template and the existing resource had to have exactly matching labels and annotations. But in some cases, external controllers might add a new label or annotation, which this controller would see as a mismatch. Now it just ensures that the desired labels and annotations exist and match.

Refs:
 - https://issues.redhat.com/browse/ACM-21219


dhaiducek (approved):
LGTM! I was wondering whether we wanted to create an exception for `gatekeeper.sh` annotations, but this makes sense and would prevent churn if a label or annotation we don't expect appears and causes the same issue.

PR #166, opened 2025-07-17
Fix restarting the gatekeeper-constraint-sync

JustinKuli:
Previously, when Gatekeeper was uninstalled, the context for the gatekeeper sync manager was permanently cancelled. As a result, if Gatekeeper was re-installed, the manager would immediately stop again, and the constraint-sync controlelr would not be started properly.

Refs:
 - https://issues.redhat.com/browse/ACM-21861

### DIFF
--- a/controllers/gatekeepersync/gatekeeper_constraint_sync.go
+++ b/controllers/gatekeepersync/gatekeeper_constraint_sync.go
@@ -50,10 +50,15 @@ var (
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *GatekeeperConstraintReconciler) SetupWithManager(mgr ctrl.Manager, constraintEvents source.Source) error {
+	skipNameValidation := true // we need to be able to stop and restart this controller
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&policyv1.Policy{}).
 		WithEventFilter(policyPredicates()).
-		WithOptions(controller.Options{MaxConcurrentReconciles: r.ConcurrentReconciles}).
+		WithOptions(controller.Options{
+			SkipNameValidation:      &skipNameValidation,
+			MaxConcurrentReconciles: r.ConcurrentReconciles,
+		}).
 		WatchesRawSource(constraintEvents).
 		Named(ControllerName).
 		Complete(r)
--- a/main.go
+++ b/main.go
@@ -955,6 +955,11 @@ func manageGatekeeperSyncManager(
 			mgrRunning = false
 
 			mgrCtxCancel()
+
+			// Reset the context for later, otherwise the context is permanently cancelled,
+			// and the manager won't start if Gatekeeper is reinstalled.
+			//nolint:fatcontext
+			mgrCtx, mgrCtxCancel = context.WithCancel(ctx)
 		}
 
 		select {
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -12,6 +12,7 @@ import (
 	"os/user"
 	"path/filepath"
 	"testing"
+	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -45,6 +46,7 @@ var (
 	gvrEvent               schema.GroupVersionResource
 	gvrConfigurationPolicy schema.GroupVersionResource
 	gvrConstraintTemplate  schema.GroupVersionResource
+	gvrCRD                 schema.GroupVersionResource
 	kubeconfigHub          string
 	kubeconfigManaged      string
 	defaultTimeoutSeconds  int
@@ -106,6 +108,12 @@ var _ = BeforeSuite(func() {
 		Version:  "v1",
 		Resource: "constrainttemplates",
 	}
+	gvrCRD = schema.GroupVersionResource{
+		Group:    "apiextensions.k8s.io",
+		Version:  "v1",
+		Resource: "customresourcedefinitions",
+	}
+
 	clientHub = NewKubeClient("", kubeconfigHub, "")
 	clientHubDynamic = NewKubeClientDynamic("", kubeconfigHub, "")
 	clientManaged = NewKubeClient("", kubeconfigManaged, "")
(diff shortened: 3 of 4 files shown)


main.go
```diff
@@ -955,6 +955,11 @@ func manageGatekeeperSyncManager(
 			mgrRunning = false
 
 			mgrCtxCancel()
+
+			// Reset the context for later, otherwise the context is permanently cancelled,
+			// and the manager won't start if Gatekeeper is reinstalled.
+			//nolint:fatcontext
+			mgrCtx, mgrCtxCancel = context.WithCancel(ctx)
```
dhaiducek, line 962: I'm not super up on controller-runtime (but really, who is?!). If we get to this point where `mgrCtx` has been closed, is it necessary to also call cancel on it in the first place? Maybe all of this code can just be removed? I guess I'm not 100% sure what calling cancel would do here.

controllers/gatekeepersync/gatekeeper_constraint_sync.go
```diff
@@ -50,10 +50,15 @@ var (
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *GatekeeperConstraintReconciler) SetupWithManager(mgr ctrl.Manager, constraintEvents source.Source) error {
+	skipNameValidation := true // we need to be able to stop and restart this controller
```
dhaiducek, line 53: Oh yeah. I dealt with this one for the gatekeeper operator. A bit silly TBH.

test/e2e/e2e_suite_test.go
```diff
@@ -174,6 +182,45 @@ var _ = BeforeSuite(func() {
 		ClientSet:        kubernetes.NewForConfigOrDie(managedConfig),
 		ControllerName:   "status-sync-controller-test",
 	}
+
+	if !gkSyncDisabled {
```
dhaiducek, line 186: So this shuts down the controller and restarts it even before the tests are run? I guess that works... 😅

test/e2e/e2e_suite_test.go
```diff
@@ -174,6 +182,45 @@ var _ = BeforeSuite(func() {
...
+		unstructured.RemoveNestedField(originalCRD.Object, "metadata", "uid")
+		unstructured.RemoveNestedField(originalCRD.Object, "metadata", "generation")
+		unstructured.RemoveNestedField(originalCRD.Object, "metadata", "creationTimestamp")
+		unstructured.RemoveNestedField(originalCRD.Object, "status")
+
+		By("Re-creating the constrainttemplate CRD to simulate re-installing Gatekeeper")
+		_, err = clientManagedDynamic.Resource(gvrCRD).Create(context.TODO(), originalCRD, metav1.CreateOptions{})
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Waiting 10 seconds for the manager to detect the fresh CRD")
```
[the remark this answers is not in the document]
dhaiducek: ...because the tests pass? 🤷‍♂️ 😅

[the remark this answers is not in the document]
JustinKuli: Yeah - there are tests that verify the manager is working.

I had checked that without the fresh `mgrCtx` change, the tests fail.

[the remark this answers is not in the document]
yiraeChristineKim: um... Should we check log? I had a similar test. I save all logs in background. I was checking log https://github.com/stolostron/mtv-integrations but Is there any good idea?

main.go
JustinKuli: Calling `mgrCtxCancel()` is what stops the manager. The manager can also stop if the parent `ctx` is cancelled, but in that case I think *everything* in the container is shutting down.

There's some possibly confusing re-naming in the block above:
```
// Keep retrying to start the Gatekeeper sync manager until mgrCtx closes.
go func(ctx context.Context) {
  for {
    select {
    case <-ctx.Done():
      wg.Done()

      return
    default:
      log.Info(
        "Gatekeeper is installed. Starting the gatekeeper-constraint-status-sync controller.",
      )

      err := runGatekeeperSyncManager(ctx, managedCfg, mgrOptions)
      // The error is logged in runGatekeeperSyncManager since it has more context.
      if err != nil {
        time.Sleep(time.Second)
      }
    }
  }
}(mgrCtx)
```

So the `mgrCtx` we're cancelling is the one currently used by `runGatekeeperSyncManager`. We need to make a *new* context, so that in the future, if we need to run that manager again, it's not already cancelled.

test/e2e/e2e_suite_test.go
JustinKuli: Yeah, because there were two `Describes` that seemed to test some Gatekeeper stuff, I decided to do the fake uninstall and reinstall in the BeforeSuite, when I know that process won't interfere with any running tests.

test/e2e/e2e_suite_test.go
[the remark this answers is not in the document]
yiraeChristineKim: Can we add comment something like 
"Subsequent tests should run correctly even if the CRD is recreated"?
 "The controller should function properly in the following tests, even after the CRD is recreated"?
 "These tests should pass regardless of whether the CRD has been recreated, confirming controller stability."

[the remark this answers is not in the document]
JustinKuli: Comment added!

test/e2e/e2e_suite_test.go
```diff
@@ -174,6 +182,50 @@ var _ = BeforeSuite(func() {
 		ClientSet:        kubernetes.NewForConfigOrDie(managedConfig),
 		ControllerName:   "status-sync-controller-test",
 	}
+
+	if !gkSyncDisabled {
+		// This section deletes and then re-creates the Gatekeeper ConstraintTemplate CRD, which
+		// should cause the gatekeeper-sync controller to stop and then restart. Tests then verify
```
yiraeChristineKim, line 188: Can you add which tests verify that the gatekeeper-sync controller is running

dhaiducek (approved):
LGTM!

PR #230, opened 2026-06-19
fix: check ownerRef before adding status history

jan-law:
ref: https://redhat.atlassian.net/browse/ACM-34512

When two policy templates share the same name, validate the template belongs to the Policy before appending its history to the Policy

Depends on https://github.com/open-cluster-management-io/governance-policy-framework-addon/pull/229

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved policy template history tracking to prevent events from unrelated policies from appearing in managed policy status.
  * Correctly handles missing or unavailable templates during status synchronization.
  * Ensures stale foreign history is removed during subsequent reconciliations.

* **Tests**
  * Added coverage for template ownership detection and compliance-history extraction.
  * Expanded end-to-end validation for duplicate templates and isolated status history.


JustinKuli (approved):
LGTM!

PR #239, opened 2026-07-29
Retry manager creation on startup failures

JustinKuli:
Sometimes when the controller-runtime manager fails to be created, the process will exit almost immediately, which can confuse CRI-O/kubelet. This can cause the container to take longer to recover than necessary.

Now, manager creation will always be retried once (after 2 seconds), and errors that are likely to be transient connectivity problems will have additional retries after 4, 8, and 16 additional seconds. After that, the container will fully fail and kubelet can take over the restarts.

Refs:
 - https://issues.redhat.com/browse/ACM-34590

## Summary by CodeRabbit

* **Reliability Improvements**
  * Controller managers now automatically retry startup after transient network/DNS/connection-related errors, with increasing delays.
  * Startup now respects application context cancellation and stops promptly.
* **Tests**
  * Added unit coverage for determining which startup errors are treated as retryable.
  * Expanded the test suite to include the utilities package.


dhaiducek:
There are linting failures to fix here

jan-law:
https://github.com/open-cluster-management-io/governance-policy-framework-addon/pull/238 is ready for review once the tests pass

jan-law (approved):
LGTM

